Issue #17, opened 2023-03-01
[Feature] Allow forcing feature values via URL query parameters


tinahollygb:
### Acceptance Criteria

Allow a user to pass in query parameters for feature values, as well as on/off for boolean features to force evaluations.

PR #18, opened 2023-03-03
Force feature values using the URL

tinahollygb:
Adds support for parsing feature default values from the URL for the following data types:

- boolean (as 0/1, true/false, on/off)
- integer
- float
- string
- json
    - Gson deserializable that doesn't require a custom type attribute
    - from string, allowing a custom deserialization implementation (see JSON region in tests)

Closes https://github.com/growthbook/growthbook-sdk-java/issues/17

### DIFF
--- a/lib/src/main/java/growthbook/sdk/java/GrowthBookUtils.java
+++ b/lib/src/main/java/growthbook/sdk/java/GrowthBookUtils.java
@@ -1,5 +1,7 @@
 package growthbook.sdk.java;
 
+import com.google.gson.Gson;
+
 import javax.annotation.Nullable;
 import java.math.BigInteger;
 import java.net.MalformedURLException;
@@ -292,6 +294,24 @@ class GrowthBookUtils {
     }
 
     // endregion Forced feature for URL -> Integer
+
+    // region Forced feature for URL -> Objects
+
+    @Nullable
+    public static <ValueType> ValueType getForcedSerializableValueFromUrl(String featureKey, URL url, Class<ValueType> valueTypeClass, Gson gson) {
+        String value = getForcedFeatureRawValueForKeyFromUrl(featureKey, url);
+
+        if (value == null) return null;
+
+        try {
+            return gson.fromJson(value, valueTypeClass);
+        } catch (Exception e) {
+            e.printStackTrace();
+            return null;
+        }
+    }
+
+    // endregion Forced feature for URL -> Objects
 
     // endregion Forced feature for URL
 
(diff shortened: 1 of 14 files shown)


lib/src/main/java/growthbook/sdk/java/GrowthBookUtils.java
```diff
@@ -153,6 +157,164 @@ public static Integer getQueryStringOverride(String id, URL url, Integer numberO
...
+        if (value.equals("true") || value.equals("1") || value.equals("on")) {
+            return true;
+        }
+
+        if (value.equals("false") || value.equals("0") || value.equals("off")) {
+            return false;
+        }
+
+        return null;
+    }
```
tinahollygb, line 208: Boolean support is on/off, 1/0, true/false (case insensitive).

lib/src/main/java/growthbook/sdk/java/GrowthBookUtils.java
```diff
@@ -153,6 +157,164 @@ public static Integer getQueryStringOverride(String id, URL url, Integer numberO
...
+
+    /**
+     * Evaluate a forced string value from a URL. If the provided key is not present in the URL, return null.
+     * @param featureKey    feature ID/key (not prefixed with gb~)
+     * @param url    Page URL to evaluate for forced feature values
+     * @return  value or null
+     */
+    @Nullable
+    public static String getForcedStringValueFromUrl(String featureKey, URL url) {
+        return getForcedFeatureRawValueForKeyFromUrl(featureKey, url);
```
tinahollygb, line 222: Nothing special with string evaluation, just get the raw value from the URL.

lib/src/test/java/growthbook/sdk/java/GrowthBookTest.java
```diff
@@ -527,4 +530,397 @@ void test_getFeatureValue_gsonDeserializable_nullValueUsesDefaultValue() {
...
+            .build();
+        GrowthBook subject = new GrowthBook(context);
+
+        String resultAsString = (String) subject.getFeatureValue("meal_overrides_gluten_free", "{\"meal_type\": \"standard\", \"dessert\": \"Donut\"}");
+        // Custom deserialization example
+        MealOrder result = jsonUtils.gson.fromJson(resultAsString, MealOrder.class);
+
+        assertEquals(MealType.GLUTEN_FREE, result.getMealType());
+        assertEquals("French Vanilla Ice Cream", result.getDessert());
+    }
```
tinahollygb, line 762: Here are some JSON examples

PR #22, opened 2023-04-17
New bucketing support with hash version 2, bump tests to v0.4.1

tinahollygb:
Adds support for the new bucketing algorithm with hash version 2.

Removes some custom serialization that wasn't absolutely necessary, which simplifies the code.


lib/src/main/java/growthbook/sdk/java/GBContext.java
```diff
@@ -48,14 +50,22 @@ public GBContext(
...
+        // Features start as empty JSON
+        this.featuresJson = "{}";
+        if (encryptionKey != null && featuresJson != null) {
+            // Attempt to decrypt payload
+            try {
+                String decrypted = DecryptionUtils.decrypt(featuresJson, encryptionKey);
+                this.featuresJson = decrypted.trim();
+            } catch (DecryptionUtils.DecryptionException e) {
+                e.printStackTrace();
+            }
```
tinahollygb, line 62: This should be safer at runtime now.

lib/src/main/java/growthbook/sdk/java/DecryptionUtils.java
```diff
@@ -16,9 +19,16 @@
...
 class DecryptionUtils {
-    public static String decrypt(String payload, String encryptionKey) {
+
+    public static class DecryptionException extends Exception {
+        public DecryptionException(String errorMessage) {
+            super(errorMessage);
+        }
+    }
+
+    public static String decrypt(String payload, String encryptionKey) throws DecryptionException {
```
tinahollygb, line 29: Changing the exception to `DecryptionException` so that it's checked (instead of the `IllegalArgumentException` which is a runtime error). We are handling it in the `GBContext` and `GBFeaturesRepository`.

lib/build.gradle
```diff
@@ -25,6 +25,8 @@ repositories {
 }
 
 dependencies {
+    implementation 'org.jetbrains:annotations:24.0.0'
```
tinahollygb, line 28: This helps in development but doesn't offer anything helpful at compile-time, at least out of the box.

lib/src/main/java/growthbook/sdk/java/Experiment.java
```diff
@@ -92,40 +109,7 @@ public String toString() {
...
-     * @return a serializer for {@link Experiment}
-     * @param <ValueType> value type for the Experiment
-     */
-    public static <ValueType> JsonSerializer<Experiment<ValueType>> getSerializer() {
-        return new JsonSerializer<Experiment<ValueType>>() {
-            @Override
-            public JsonElement serialize(Experiment<ValueType> src, Type typeOfSrc, JsonSerializationContext context) {
-                return Experiment.getJson(src);
-            }
-        };
```
tinahollygb, line 128: Removing the custom serialization results in the `condition` field not being present and instead `conditionJson` on the serialized JSON but we don't need that to be a specific shape for any reason, so it's fine. Removing the custom serialization, while resulting in less-than-perfect JSON output, makes it easier to maintain.

lib/src/main/java/growthbook/sdk/java/ExperimentResult.java
```diff
@@ -1,57 +1,96 @@
...
- * <li>hashValue (string) - The value of that attribute</li>
- * <li>featureId (string or null) - The id of the feature (if any) that the experiment came from</li>
- * </ul>
- *
+ * The result of an {@link GrowthBook#run(Experiment)} call
  * @param <ValueType> generic type for the value type for this experiment's variations.
  */
 @Data
-@Builder
-@AllArgsConstructor
```
tinahollygb, line 31: Moved from this property-based approach to the explicit `@Builder`-annotated constructor.

lib/src/main/java/growthbook/sdk/java/Feature.java
```diff
@@ -1,12 +1,6 @@
 package growthbook.sdk.java;
 
-import com.google.gson.JsonArray;
-import com.google.gson.JsonElement;
-import com.google.gson.JsonObject;
-import com.google.gson.reflect.TypeToken;
```
tinahollygb, line 6: All the JSON serialization stuff was not being used in the actual implementation, only in tests, so I've removed it since we convert to/from JSON with the automated serialization with Gson.

lib/src/main/java/growthbook/sdk/java/GrowthBookJsonUtils.java
```diff
@@ -38,12 +41,6 @@ private GrowthBookJsonUtils() {
         // FeatureResult
         gsonBuilder.registerTypeAdapter(FeatureResult.class, FeatureResult.getSerializer());
 
-        // Experiment
-        gsonBuilder.registerTypeAdapter(Experiment.class, Experiment.getSerializer());
-
-        // ExperimentResult
-        gsonBuilder.registerTypeAdapter(ExperimentResult.class, ExperimentResult.getSerializer());
```
tinahollygb, line 45: Went back to default serialization. It results in the `conditionJson` being printed as a property, and means we cannot check the whole object, but it'll be easier to maintain.

Issue #23, opened 2023-04-20
The SDK should use a logging library


igorlovich:
Currently the sdk uses standard out/error for logging (System.out en printStackTrace). It should use a library (slf4j tends to be the standard allowing the end user to provide the binding to the actual logging framework e.g. logback or log4j etc..) . I can provide a PR if there is an interest.

tinahollygb:
Hi @igorlovich thanks for offering to submit a PR for it. That would be great! Let's figure out the best approach. 

We are trying to keep our dependencies to a minimum to keep our distribution size down. The SDK ships to Android as well, so adding to clients' bundle size is something we're trying to avoid if we can. If it can be done in a way that does not add additional dependencies and can be supported on Android, or the added dependencies are small, I think that would be acceptable.

I think it probably makes the most sense to have it added to the `GBContext`, as the last argument so that it's supported in the context builder. The SDK should not fail if a logger is not provided.

Is the [slf4j](https://www.slf4j.org/manual.html) library you're suggesting compatible with both Android and the JVM? If it's small and simple, I think we would be open to that change, e.g. if it's only a set of interfaces.

Did you want to discuss your suggested change first before getting started? Thanks.

tinahollygb:
Also, to follow up with the previous message, these are the example projects we'd test that logging works for:

- https://github.com/growthbook/examples/tree/main/jvm-spring-web
- https://github.com/growthbook/examples/tree/main/jvm-kotlin-ktor-example
- https://github.com/growthbook/examples/tree/main/android-example

igorlovich:
I understand your concerns and will try to address them one by one

1. The -api jar is mostly interfaces and the latest version is 62 kb in size.

2. The library does not do any logging as such; it is a facade (Slf4j = Simple Logging Facade for Java). By including the appropriate binding you can use whatever logging framework you want
e.g. slf4j-logback, slf4j-log4j etc.. etc..

3. For android logging there is slf4j-android binding, or if the project is using logback / log4j then one of the specific binding can  be used. 

4. If no binding is provided  the logging operation is a no-op.

5. If the user wants to maintain the current behavior (i.e. printing to stdout/stderr) then they would need to provide sl4j-simple binding (12k in size)


Does this answer all your questions ?

tinahollygb:
> Does this answer all your questions ?

@igorlovich yes! That seems like a good solution then, thanks for explaining it.

Feel free to proceed with that approach. Thanks again for offering!

[...]

PR #24, opened 2023-04-25
GH-26 The SDK should use a logging library

igorlovich:
As per [GH-26](https://github.com/growthbook/growthbook-sdk-java/issues/23), I have migrated all the System.out and .printStackTrace to logging calls.

I noticed there are a bunch of commented out println statements in the code; We may want to un-comment them and log them at .debug or .trace levels, depending on how you use them.

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,6 +12,7 @@
 ## Requirements
 
 - Java version 1.8 or later
+- The library uses Slf4j for logging. You will need to add an appropriate slf4j binding for the logging framework that your project uses.
 
 
 ## Documentation
--- a/lib/build.gradle
+++ b/lib/build.gradle
@@ -34,6 +34,10 @@ dependencies {
     implementation 'org.apache.commons:commons-math3:3.6.1'
     implementation 'com.google.guava:guava:31.0.1-jre'
 
+    // https://mvnrepository.com/artifact/org.slf4j/slf4j-api
+    implementation 'org.slf4j:slf4j-api:2.0.7'
+    testImplementation 'org.slf4j:slf4j-simple:2.0.7'
+
     // https://mvnrepository.com/artifact/com.google.code.gson/gson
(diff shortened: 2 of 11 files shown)


tinahollygb:
These annotations look fairly magical. I understand they inject a logger at the field level, but how does this work? How does a user provide their own logger? What if a project doesn't use Slf4j and uses a different logger, how would they provide a logger? I don't see any new properties added to the GrowthBook constructor and builder so it's not clear how they would add a logger.

igorlovich:
`@Slf4j` is a lombok annotation that, as you have noticed, creates the log field; so nothing to do with slf4j. This can be done manually, without lombok, `private final Logger log = LoggerFactory.getLogger(HelloWorld.class);`. 

It is important to keep in mind that `slf4j` is NOT a logging framework. It is a facade that allows a project to use whatever framework they want. This especially key for library developers as, they do not know which logging solution this or that project uses. 

So let's take an example where the project uses `log4j2` as the logging framework. It would already have the log4j2 dependency and be configured as desired. When they add the` growthbook-sdk-java` dependency, and they want to see its logs, they would add the `log4j-slf4j2-impl` binding dependency and the logs would show up as per their configuration. 

If the project was using  `logback,` they would not event need to add any dependencies as `logback` implements the `slf4j` api directly.

As you can see, I have added the `slf4j-simple` as a test dependency. This binding just prints to stdout / stderror, so during test execution the log messages will be printed out to console, same as before.

README.md
```diff
@@ -12,6 +12,7 @@
 ## Requirements
 
 - Java version 1.8 or later
+- The library uses Slf4j for logging. You will need to add an appropriate slf4j binding for the logging framework that your project uses.
```
tinahollygb, line 15: This looks like if people don't use slf4j it'll break, is that correct?

PR #30, opened 2023-07-06
Fix isOn() for float values

### LINKED ISSUE #29: [Bug] isOn() does not return true for truthy Float values
There is a bug that when using a float value, `isOn()` is not true.

`isOn()` should return true for all non-zero float values.

tinahollygb:
fixes #29 check for instance of Double in FeatureResult#isOn()

### DIFF
--- a/lib/src/test/java/growthbook/sdk/java/FeatureResultTest.java
+++ b/lib/src/test/java/growthbook/sdk/java/FeatureResultTest.java
@@ -2,8 +2,7 @@ package growthbook.sdk.java;
 
 import org.junit.jupiter.api.Test;
 
-import static org.junit.jupiter.api.Assertions.assertEquals;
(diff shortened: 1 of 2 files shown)


lib/src/test/java/growthbook/sdk/java/FeatureResultTest.java
```diff
@@ -62,4 +61,110 @@ void featureResultSourceOutputsCorrectlyToJson() {
...
+    void featureResult_isOn_withNonZeroValue_returnsTrue_forDoubles() {
+        FeatureResult<Double> subject = FeatureResult
+            .<Double>builder()
+            .value(1.0)
+            .source(FeatureResultSource.FORCE)
+            .build();
+
+        assertTrue(subject.isOn());
+        assertFalse(subject.isOff());
+    }
```
tinahollygb, line 129: This is the test case that would fail.

PR #31, opened 2023-07-07
Feature/add error handler for callback

### LINKED ISSUE #32: Subscribe to feature refresh errors
I recently created a #31  to include a callback on error feature. I want an answer about this. I am ready to correct my pull request and add tests. Thank you)

AmirSalakhov94:
Hello. We want to use client-side Growthbook service call metrics. But a handler made for succesfully calls only would like to add an error handler as well. Here's the pull request, it's not claims to be the best solution. Thank you)

Closes #32

### DIFF
--- a/lib/src/main/java/growthbook/sdk/java/GBFeaturesRepository.java
+++ b/lib/src/main/java/growthbook/sdk/java/GBFeaturesRepository.java
@@ -98,9 +98,8 @@ public class GBFeaturesRepository implements IGBFeaturesRepository {
 
     /**
      * Subscribe to feature refresh events
-     * This callback is called when the features are successfully refreshed.
+     * This callback is called when the features are successfully or error refreshed.
      * This is called even if the features have not changed.
-     * This will not be called if fetching the features results in a failure.
      * @param callback  This callback will be called when features are refreshed
      */
     @Override
@@ -124,6 +123,7 @@ public class GBFeaturesRepository implements IGBFeaturesRepository {
             @Override
             public void onFailure(@NotNull Call call, @NotNull IOException e) {
                 // OkHttp will auto-retry on failure
+                self.refreshFailedCallbacks(e);
             }
 
             @Override
@@ -233,9 +233,8 @@ public class GBFeaturesRepository implements IGBFeaturesRepository {
 
             this.featuresJson = refreshedFeatures;
 
-            this.refreshCallbacks.forEach(featureRefreshCallback -> {
-                featureRefreshCallback.onRefresh(this.featuresJson);
-            });
+            this.refreshSuccessfullyCallbacks(this.featuresJson);
+
         } catch (IOException | DecryptionUtils.DecryptionException e) {
             e.printStackTrace();
 
@@ -246,7 +245,15 @@ public class GBFeaturesRepository implements IGBFeaturesRepository {
         }
     }
 
+    private void refreshSuccessfullyCallbacks(String featuresJson) {
+        this.refreshCallbacks
+                .forEach(featureRefreshCallback -> featureRefreshCallback.onRefresh(featuresJson));
+    }
 
+    private void refreshFailedCallbacks(Throwable throwable) {
+        this.refreshCallbacks
+                .forEach(featureRefreshCallback -> featureRefreshCallback.onError(throwable));
+    }
 
     /**
      * Appends User-Agent info to the request headers.
--- a/lib/src/main/java/growthbook/sdk/java/GBFeaturesRepository.java
+++ b/lib/src/main/java/growthbook/sdk/java/GBFeaturesRepository.java
@@ -123,7 +123,7 @@ public class GBFeaturesRepository implements IGBFeaturesRepository {
             @Override
             public void onFailure(@NotNull Call call, @NotNull IOException e) {
                 // OkHttp will auto-retry on failure
-                self.refreshFailedCallbacks(e);
+                self.onRefreshFailed(e);
             }
 
             @Override
@@ -233,8 +233,7 @@ public class GBFeaturesRepository implements IGBFeaturesRepository {
 
             this.featuresJson = refreshedFeatures;
 
-            this.refreshSuccessfullyCallbacks(this.featuresJson);
-
+            this.onRefreshSuccess(this.featuresJson);
         } catch (IOException | DecryptionUtils.DecryptionException e) {
             e.printStackTrace();
 
@@ -245,14 +244,16 @@ public class GBFeaturesRepository implements IGBFeaturesRepository {
         }
     }
 
-    private void refreshSuccessfullyCallbacks(String featuresJson) {
-        this.refreshCallbacks
-                .forEach(featureRefreshCallback -> featureRefreshCallback.onRefresh(featuresJson));
+    private void onRefreshSuccess(String featuresJson) {
+        this.refreshCallbacks.forEach(featureRefreshCallback -> {
+            featureRefreshCallback.onRefresh(featuresJson);
+        });
     }
 
-    private void refreshFailedCallbacks(Throwable throwable) {
-        this.refreshCallbacks
-                .forEach(featureRefreshCallback -> featureRefreshCallback.onError(throwable));
+    private void onRefreshFailed(Throwable throwable) {
+        this.refreshCallbacks.forEach(featureRefreshCallback -> {
+            featureRefreshCallback.onError(throwable);
+        });
     }
 
     /**
--- a/lib/src/main/java/growthbook/sdk/java/GBFeaturesRepository.java
+++ b/lib/src/main/java/growthbook/sdk/java/GBFeaturesRepository.java
@@ -98,7 +98,7 @@ public class GBFeaturesRepository implements IGBFeaturesRepository {
 
     /**
      * Subscribe to feature refresh events
-     * This callback is called when the features are successfully or error refreshed.
+     * This callback is called when the features are successfully refreshed or there is an error when refreshing.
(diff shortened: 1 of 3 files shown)


lib/src/main/java/growthbook/sdk/java/GBFeaturesRepository.java
```diff
@@ -98,9 +98,8 @@ public String getFeaturesJson() {
 
     /**
      * Subscribe to feature refresh events
-     * This callback is called when the features are successfully refreshed.
+     * This callback is called when the features are successfully or error refreshed.
```
tinahollygb, line 101:
```suggestion
     * This callback is called when the features are successfully refreshed or there is an error when refreshing.
```

lib/src/main/java/growthbook/sdk/java/GBFeaturesRepository.java
```diff
@@ -246,7 +245,15 @@ private void onSuccess(Response response) throws FeatureFetchException {
         }
     }
 
+    private void refreshSuccessfullyCallbacks(String featuresJson) {
```
tinahollygb, line 248: Can you rename this to `onRefreshSuccess(String featuresJson)`

lib/src/main/java/growthbook/sdk/java/GBFeaturesRepository.java
```diff
@@ -246,7 +245,15 @@ private void onSuccess(Response response) throws FeatureFetchException {
...
+        this.refreshCallbacks
+                .forEach(featureRefreshCallback -> featureRefreshCallback.onRefresh(featuresJson));
+    }
 
+    private void refreshFailedCallbacks(Throwable throwable) {
```
tinahollygb, line 253: Can you rename this to `onRefreshFailed(Throwable throwable)`

lib/src/main/java/growthbook/sdk/java/GBFeaturesRepository.java
```diff
@@ -233,9 +233,8 @@ private void onSuccess(Response response) throws FeatureFetchException {
 
             this.featuresJson = refreshedFeatures;
 
-            this.refreshCallbacks.forEach(featureRefreshCallback -> {
-                featureRefreshCallback.onRefresh(this.featuresJson);
-            });
+            this.refreshSuccessfullyCallbacks(this.featuresJson);
+
```
tinahollygb, line 237: Style: Unnecessary whitespace

lib/src/main/java/growthbook/sdk/java/GBFeaturesRepository.java
```diff
@@ -246,7 +245,15 @@ private void onSuccess(Response response) throws FeatureFetchException {
         }
     }
 
+    private void refreshSuccessfullyCallbacks(String featuresJson) {
+        this.refreshCallbacks
+                .forEach(featureRefreshCallback -> featureRefreshCallback.onRefresh(featuresJson));
```
tinahollygb, line 250: Style: Can you use the existing style (add braces).

lib/src/main/java/growthbook/sdk/java/GBFeaturesRepository.java
```diff
@@ -246,7 +245,15 @@ private void onSuccess(Response response) throws FeatureFetchException {
...
+    }
 
+    private void refreshFailedCallbacks(Throwable throwable) {
+        this.refreshCallbacks
+                .forEach(featureRefreshCallback -> featureRefreshCallback.onError(throwable));
```
tinahollygb, line 255: Same style request here as above re: braces

tinahollygb (changes requested):
Functionality-wise seems good, but I left some grammar and style requests. See inline. Thanks!

lib/src/main/java/growthbook/sdk/java/GBFeaturesRepository.java
AmirSalakhov94: corrected

lib/src/main/java/growthbook/sdk/java/GBFeaturesRepository.java
AmirSalakhov94: corrected

lib/src/main/java/growthbook/sdk/java/GBFeaturesRepository.java
AmirSalakhov94: corrected)

lib/src/main/java/growthbook/sdk/java/GBFeaturesRepository.java
AmirSalakhov94: corrected)

AmirSalakhov94:
> Functionality-wise seems good, but I left some grammar and style requests. See inline. Thanks!

Thank you, I was glad to contribute) I added 2 more tests to check refresh callback.

lib/src/test/java/growthbook/sdk/java/GBFeaturesRepositoryTest.java
```diff
@@ -137,8 +139,65 @@ void canFetchEncryptedFeatures_mockedResponse() throws IOException, FeatureFetch
...
+            public void onRefresh(String featuresJson) {
+            }
 
+            @Override
+            public void onError(Throwable throwable) {
+                assertThrows(IOException.class, () -> {
+                    throw throwable;
+                });
+            }
+        });
```
tinahollygb, line 197: I don't think this assertion is working as expected. I'm assuming you're trying to test that the onError callback is called when it fails, but when I comment out the real code, this test still passes, which leads me to believe it isn't testing anything.

AmirSalakhov94: I slightly corrected the test, if I understood you correctly, now when you comment on the real code, it ends with an error. If you have any ideas how it can be implemented, I will be grateful)

tinahollygb: There is some strange behaviour with the tests still. I notice that the `getResponse` method you've abstracted from the initial success is used for the failure, and it responds with a 200. I see that when I change the failed response JSON, it fails because it tries to successfully process the failed JSON response data. It looks like both the success and failure cases are being called in the tests. I don't think the failure tests are accurately being tested—we are testing a condition that would never occur (both a success and a failure). You can see the odd behaviour by changing `testOnFeaturesRefresh_Error()`'s fake response to be `"{}"`—we get: `CONFIGURATION_ERROR : No features found` which means it's trying to run the success.

tinahollygb (changes requested):
See inline comment but tests are still not accurately testing a real failure case scenario.

lib/src/test/java/growthbook/sdk/java/GBFeaturesRepositoryTest.java
AmirSalakhov94: oh yeah i see...it's happening because i'm calling the subject.initialize() method, it's not really related to the error check as it doesn't call onRefreshFailed, i can remove it from the call and leave only subject.getFeaturesJson(), then the test will work correctly

lib/src/test/java/growthbook/sdk/java/GBFeaturesRepositoryTest.java
```diff
@@ -158,19 +206,54 @@ private static OkHttpClient mockHttpClient(final String serializedBody) throws I
...
+    /**
+     * Create a mock instance for failed callback of {@link OkHttpClient}
+     *
+     * @param serializedBody JSON string response
+     * @return mock {@link OkHttpClient}
+     */
+    private static OkHttpClient mockHttpClientFailedCallback(final String serializedBody, final IOException ioException) throws IOException {
+        OkHttpClient okHttpClient = mock(OkHttpClient.class);
+        Call remoteCall = mock(Call.class);
+        when(remoteCall.execute()).thenReturn(getResponse(serializedBody));
```
tinahollygb, line 230: It might make sense to pass the status code into `getResponse`—it would make more sense to say that a failure would return an error code instead of 200.

Alternatively, don't abstract `getResponse` (that'll probably make it easier to read) and just do a new failure response inline where an error status is used (and not 200).

AmirSalakhov94: @tinahollygb hi) so, I rolled back my changes in the test that affects your implementation. Аdded for test testOnFeaturesRefresh_Error failed response.

lib/src/test/java/growthbook/sdk/java/GBFeaturesRepositoryTest.java
```diff
@@ -159,14 +234,14 @@ private static OkHttpClient mockHttpClient(final String serializedBody) throws I
...
-                ))
-            .build();
+                .request(new Request.Builder().url("http://url.com").build())
+                .protocol(Protocol.HTTP_1_1)
+                .code(200).message("").body(
+                        ResponseBody.create(
+                                serializedBody,
+                                MediaType.parse("application/json")
+                        ))
+                .build();
```
tinahollygb, line 244: Can you clean up these unnecessary changes?

AmirSalakhov94: sorry it's because of my auto style code

tinahollygb (approved):

tinahollygb:
Awesome, thanks for the contribution! 🚀

AmirSalakhov94:
> Awesome, thanks for the contribution! 🚀

Glad to take part)

PR #35, opened 2023-07-19
Improve `evalFeature` performance: generate fewer `MalformedURLException`

### LINKED ISSUE #34: Spend less time generating `MalformedUrlException`
# Goal

To increase performance of `evalFeature`.

# Summary

Profiling our application revealed that growthbook spent a lot of time in the URL constructor when `evalFeature` was called.

# Steps to reproduce

In our application, we set `featuresJson` on the GBContext builder and leave the URL unset.

Repeated calls to `evalFeature` results in a noticeable amount of time spent in the URL constructor @ `GrowthBookUtils/getQueryStringOverride`.

# Resolution

Growthbook should check if `urlString` is null or empty before entering the try/catch block in `GrowthBookUtils/getQueryStringOverride`.

crinklywrappr:
Resolve #34


tinahollygb (approved):
Awesome, thanks!

PR #36, opened 2023-08-01
Server-Sent Events (SSE) support 

tinahollygb:
Adds support for server-sent events.

A new class to help test with has been added, `SSETestServer`. Run the `main` to run the server. You can change the config to your own values and make changes to see it update immediately (after you refresh the /ping) endpoint.

- Falls back to SWR refresh strategy (original, default) when SSE not supported
- Reconnects after the SSE server disconnects us (every 60 seconds) and whenever the connection is closed (e.g. networking error)

Breaking changes:

- `GBFeaturesRepository` constructor signature and builder methods have changed to split the previous `endpoint` into `clientKey` and `apiHost`.


lib/src/main/java/growthbook/sdk/java/GBFeaturesRepositoryRequestInterceptor.java
```diff
@@ -0,0 +1,24 @@
...
+import okhttp3.Request;
+import okhttp3.Response;
+import org.jetbrains.annotations.NotNull;
+
+import java.io.IOException;
+
+/**
+ * Appends User-Agent info to the request headers.
+ */
+public class GBFeaturesRepositoryRequestInterceptor implements Interceptor {
```
tinahollygb, line 13: This is not new, it was just moved to a separate file.

PR #40, opened 2023-08-30
Gracefully handle error response codes on features repository initialization

micpalmia:
At the moment, HTTP status codes different than 200 are completely ignored. This leads to confusing error messages that don't help clients debug the issue quickly. Adding a check for the return status code helps ensure a more meaningful error message can be returned.

Ideally, the body of the response should be logged rather than provided in the exception message. In the absence of a logging framework in the library however, I think emitting the body in the message can help debugging.

Happy to adapt as needed if you have any feedback.

### DIFF
--- a/lib/src/main/java/growthbook/sdk/java/GBFeaturesRepository.java
+++ b/lib/src/main/java/growthbook/sdk/java/GBFeaturesRepository.java
@@ -382,7 +382,13 @@ public class GBFeaturesRepository implements IGBFeaturesRepository {
             ResponseBody responseBody = response.body();
             if (responseBody == null) {
                 throw new FeatureFetchException(
-                    FeatureFetchException.FeatureFetchErrorCode.NO_RESPONSE_ERROR
+                        FeatureFetchException.FeatureFetchErrorCode.NO_RESPONSE_ERROR
+                );
+            }
+            if (response.code() != 200) {
+                throw new FeatureFetchException(
+                        FeatureFetchException.FeatureFetchErrorCode.UNKNOWN,
+                        "Unexpected HTTP response code: " + response.code() + ", body: " + responseBody.string()
                 );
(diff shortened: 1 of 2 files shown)


lib/src/main/java/growthbook/sdk/java/GBFeaturesRepository.java
```diff
@@ -382,7 +382,13 @@ private void onSuccess(Response response) throws FeatureFetchException {
             ResponseBody responseBody = response.body();
             if (responseBody == null) {
                 throw new FeatureFetchException(
-                    FeatureFetchException.FeatureFetchErrorCode.NO_RESPONSE_ERROR
+                        FeatureFetchException.FeatureFetchErrorCode.NO_RESPONSE_ERROR
+                );
+            }
+            if (response.code() != 200) {
```
tinahollygb, line 388: Hi @micpalmia thanks for your contribution. Unfortunately this approach will be quite fragile and will introduce a regression. In some cases, the API returns 304 with the content, and in those cases, this will break the SDK.

image

Please update your error check to include only error-related status codes.

tinahollygb (changes requested):
Please see inline comment re: regression.

lib/src/main/java/growthbook/sdk/java/GBFeaturesRepository.java
```diff
@@ -382,7 +382,13 @@ private void onSuccess(Response response) throws FeatureFetchException {
...
+                throw new FeatureFetchException(
```
Bohdan-Kim, line 389: Seems like if user of the SDK doesn't catch `FeatureFetchException` exception in runtime will appear.

Bohdan-Kim:
I agree, the body of the response should be logged rather than provided in the exception message.

PR #43, opened 2024-03-18
fix connection leak on ok http client

JonathanE-BWR:
fix connection leak on ok http client, move response body into try-with-resource block to auto-close in all cases


vu-linh:
@tinahollygb Hi, can you please implement this fix and ship with the next release? Our projects are making use of GrowthBook SDK and we also see the leaked connection warning on our system. Pretty please? 🙏

vulinh64:
@vazarkevych  A question: Why was this pull request closed? The connection leak problem still persisted

PR #45, opened 2024-03-27
remove dependency on bigint for hashing

gaurav-rana:


lib/src/main/java/growthbook/sdk/java/GrowthBookUtils.java
```diff
@@ -42,23 +42,19 @@ class GrowthBookUtils {
     }
 
     private static Float hashV1(String stringValue, String seed) {
-        BigInteger bigInt = MathUtils.fnv1a_32((stringValue + seed).getBytes());
-        BigInteger thousand = new BigInteger("1000");
-        BigInteger remainder = bigInt.remainder(thousand);
-
-        float remainderAsFloat = Float.parseFloat(remainder.toString());
-        return remainderAsFloat / 1000f;
+        int bigInt = MathUtils.fnv1a_32((stringValue + seed).getBytes());
```
Bohdan-Kim, line 45: Is `int` type is enough for this? Do we need more capacity like for example using `long` type instead?

PR #46, opened 2024-04-12
Add check & unit tests in case the feature value is a collection impl

vulinh64:


Bohdan-Kim:
It seems like there wasn't handled array case and this pull request handled it.

PR #48, opened 2024-04-19
Reduce Exceptions in the ConditionEvaluator

nnusse:
Hello,

this PR should reduce the amount of exceptions that the `ConditionEvaluator` would throw, if it receives incoming null values for the evaluation. 
There are some sections in the code that are not able to handle that properly. I received already the following exceptions (in the form of stack traces on the stderr) in one of my applications:

- **java.lang.IllegalStateException: Not a JSON Primitive: null**
- **java.lang.UnsupportedOperationException: JsonNull**

The SDK is catching all exceptions and additionally prints the stacktrace to the standard error output channel. Unfortunately this can also harm consumers of this library, because some applications might require a specific format of messages that get printed to the error or normal output channels. From my point of view i think it is really worth it to work on the issue (#23), but i think as well, that it is always good to handle errors properly.

I also adjusted the test to let them fail, if exceptions will occur that are not expected to appear under a specific test condition. I'm not really happy with the changes that i had to introduce in the tests, but i couldn't come up with a better solution, unfortunately the SDK is catching every error which makes it hard to check if a specific error occurred. 

I hope that you see value in this PR. 

All the best,
Norman

### DIFF
--- a/lib/src/test/resources/test-cases.json
+++ b/lib/src/test/resources/test-cases.json
@@ -1953,7 +1953,7 @@
       "$nor - pass with null value for age",
       {"$nor": [{"name": "john"}, {"age": {"$lt": 30}}]},
       {"name": "jim", "age": null},
-      true
+      false
     ],
     [
       "$nor - fail both",
--- a/lib/src/test/resources/test-cases.json
+++ b/lib/src/test/resources/test-cases.json
@@ -1372,6 +1372,7 @@
     ["$in - pass", {"num": {"$in": [1, 2, 3]}}, {"num": 2}, true],
     ["$in - fail", {"num": {"$in": [1, 2, 3]}}, {"num": 4}, false],
     ["$in - not array", {"num": {"$in": 1}}, {"num": 1}, false],
+    ["$in - fail with null value", {"num": {"$in": [1, 2, 3]}}, {"num": null}, false],
     [
       "$in - array pass 1",
       {"tags": {"$in": ["a", "b"]}},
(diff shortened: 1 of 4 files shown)


Bohdan-Kim:
Hello @nnusse!
There was a merge conflict with release/0.93 branch. 
The test `ConditionEvaluatorTest#test_evaluateCondition_testCases()` is failing after resolving merge conflict.

lib/src/test/resources/test-cases.json
```diff
@@ -1847,12 +1931,30 @@
...
+      "$nor - pass with null value for name",
+      {"$nor": [{"name": "john"}, {"age": {"$lt": 30}}]},
+      {"name": null, "age": 40},
+      true
+    ],
+    [
+      "$nor - pass with null value for age",
+      {"$nor": [{"name": "john"}, {"age": {"$lt": 30}}]},
+      {"name": "jim", "age": null},
+      true
```
Bohdan-Kim, line 1956: `null` is `0` in JavaScript.
`(0 < 30) = true`.
`false nor true = !(false or true) = false`

lib/src/test/resources/test-cases.json
```diff
@@ -1720,12 +1804,30 @@
...
+    ],
```
Bohdan-Kim, line 1957: Why it should pass?
If we follow Javascript Style `null` is `0` then `(0 < 30) = true`
`false nor true = not(false or true) = not(true) = false`

nnusse: Hey @Bohdan-Kim, thank for reviewing this PR. Unfortunately I'm ill at the moment. I will check all your comments and update the PR accordingly next week.

Bohdan-Kim: Hello @nnusse, I hope you doing well. 
Do you mind if we fix failing test by ourself in order to merge the request faster?

nnusse: You are right. I had did a mistake in my implementation that you corrected at the point where you resolved the merge conflict. I will adjust the test definition.

nnusse: I will mark that discussion as resolved. Feel free to give me more feedback if you have any further concerns or questions.

Bohdan-Kim:
It looks good

PR #58, opened 2024-05-20
Logging

vazarkevych:
printStackTrace() was replaced with logging


igorlovich:
I believe it is better to use Slf4j (as proposed in https://github.com/growthbook/growthbook-sdk-java/pull/24) so that users can pick whatever logging backend they want (including java logging)

nnusse:
I can only support @igorlovich. The SDK is now highly coupled to the java based logging framework. That would mean that every consumer/user of the SDK has to provide additional configuration to change the message format or other properties, although they might have these configuration already in place - but for the logging framework that they have in use, like Log4j or Logback. Using SLF4J would avoid such cases.

Bohdan-Kim:
So, we need to add SLF4J

PR #76, opened 2024-06-25
Add support for setting JSON Feature Object in GBContext builder

gaurav-rana:
The current GBContext implementation is suboptimal for java backend integrations.
We have to create a new GBContext for each request, and because the CustomGbContextBuilder only supports StringType featureJson, we have to pass featureJson as string and it parses the JSON (converting from String to JSONType) for each and every request.
This can be avoided by letting the application maintain JSON and set JSON directly.

Flame graph with the current GBContext:

Screenshot 2024-06-25 at 1 53 59 PM

As we can see, around 42% of total CPU is spent inside parsing this JSON for each and every request, this overhead can be clearly avoided by letting the client set the JSON directly instead of string.

After the fix:
Screenshot 2024-06-25 at 8 41 37 PM

The stack frames for transformFeatures went away.

### DIFF
--- a/lib/src/main/java/growthbook/sdk/java/GBContext.java
+++ b/lib/src/main/java/growthbook/sdk/java/GBContext.java
@@ -94,7 +94,6 @@ public class GBContext {
      * Feature definitions - To be pulled from API / Cache
      */
     @Nullable
-    @Getter(AccessLevel.PACKAGE)
     private JsonObject features;
 
     private void setFeatures(@Nullable JsonObject features) {
--- a/lib/src/main/java/growthbook/sdk/java/GBContext.java
+++ b/lib/src/main/java/growthbook/sdk/java/GBContext.java
@@ -28,6 +28,7 @@ public class GBContext {
      * Alternatively, you can use this static method instead of the builder.
      *
      * @param attributesJson                   User attributes as JSON string
+     * @param featuresJson                     Features response as JSON string, or the encrypted payload. Encrypted payload requires `encryptionKey`
      * @param features                         Features response as JSON Object, either set this or `featuresJson`
      * @param encryptionKey                    Optional encryption key. If this is not null, featuresJson should be an encrypted payload.
      * @param enabled                          Whether globally all experiments are enabled (default: true)
--- a/lib/src/main/java/growthbook/sdk/java/GBContext.java
+++ b/lib/src/main/java/growthbook/sdk/java/GBContext.java
@@ -63,7 +63,7 @@ public class GBContext {
         this.attributesJson = attributesJson == null ? "{}" : attributesJson;
 
         if (featuresJson != null) {
-            this.features = convertStringToJsonObject(featuresJson, encryptionKey);
+            this.features = transformEncryptedFeatures(featuresJson, encryptionKey);
         }
         if (features != null) {
             this.features = features;
@@ -221,7 +221,17 @@ public class GBContext {
         }
     }
 
-    public static JsonObject convertStringToJsonObject(
+    @Nullable
+    public static JsonObject transformFeatures(String featuresJsonString) {
+        try {
+            return GrowthBookJsonUtils.getInstance().gson.fromJson(featuresJsonString, JsonObject.class);
+        } catch (Exception e) {
+            log.error(e.getMessage(), e);
+            return null;
+        }
+    }
+
+    public static JsonObject transformEncryptedFeatures(
             @Nullable String featuresJson,
             @Nullable String encryptionKey
     ) {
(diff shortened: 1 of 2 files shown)


lib/src/main/java/growthbook/sdk/java/GBContext.java
```diff
@@ -94,13 +98,8 @@ public GBContext(
      * Feature definitions - To be pulled from API / Cache
      */
     @Nullable
-    @Getter(AccessLevel.PACKAGE)
     private JsonObject features;
 
-    private void setFeatures(@Nullable JsonObject features) {
-        this.features = features;
-    }
-
```
Bohdan-Kim, line 103: Could you please clarify the reason of removing setter?

gaurav-rana: it would be taken care by @Data lombok annotation
the current setter is private, we need a public setter which @Data would automatically create.

lib/src/main/java/growthbook/sdk/java/GBContext.java
```diff
@@ -46,6 +47,7 @@ public class GBContext {
     public GBContext(
             @Nullable String attributesJson,
             @Nullable String featuresJson,
+            @Nullable JsonObject features,
```
Bohdan-Kim, line 50: Lines 49 and 50 look like the user of the library should pass `String` and `JsonObject`.
Seems like two constructors needed one with `String` and another one with `JsonObject`.
Do you allow us to try to add second constructor?

gaurav-rana: yes, that makes sense.

gaurav-rana: we might have to handle @Builder annotation given that works only on one constructor.

vazarkevych: It seems like tests were passed.

Bohdan-Kim: If I remove `@Builder` annotation from the second constructor would it help?

gaurav-rana: Yes, that would be better.
The primary interface should be the one with JsonObject.

lib/src/main/java/growthbook/sdk/java/GBContext.java
```diff
@@ -235,17 +257,37 @@ static class CustomGBContextBuilder extends GBContextBuilder {
...
+            }
+        } else if (featuresJson != null) {
+            jsonObject = GBContext.transformFeatures(featuresJson);
+        }
+
+        return jsonObject;
+    }
+
     @Nullable
     private static JsonObject transformFeatures(String featuresJsonString) {
```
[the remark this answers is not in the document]
Bohdan-Kim: Ok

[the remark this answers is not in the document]
gaurav-rana: Great! i've changed `convertStringToJsonObject` from private to public.

lib/src/main/java/growthbook/sdk/java/GBContext.java
Bohdan-Kim: But without `featuresJson` typed with `String` we will need to correct all tests:
image

gaurav-rana:
```
JsonObject features = gson.fromJson(featuresJson, JsonObject.class);
GBContext context1 = GBContext.builder()
                            .features(features)
                            .attributesJson(getUserAttributesObj().toString()) // Optional
                            .trackingCallback(null)
                            .build();
GrowthBook growthBook = new GrowthBook(context1);
```
Right now with the code changes in the PR, the above way of initializing GB doesn't work, it leads to empty features jsonobject.

if we decide to not change the builder interface then we should remove `@Builder` from second constructor because it's allowing `features` to be set using the builder pattern but not using the value.
In case, anyone wants to pass features Json directly (instead of string) then either they'll have to use the constructor without builder or use the other builder and then pass feature Json separately.
Like this
```
 GBContext context1 = GBContext.builder()
                            .attributesJson(getUserAttributesObj().toString()) // Optional
                            .trackingCallback(null)
                            .build();
context1.setFeatures(features);
GrowthBook growthBook = new GrowthBook(context1);
```

Bohdan-Kim: Seems like our intention to create the second constructor is wrong because it is not pure Java constructor but constructor annotated with Lombok Builder.

gaurav-rana: The current changes looks good to me
let me know when can we merge this PR

PR #85, opened 2024-07-08
Optimize Json Conversion in ConditionEvaluator and Update GBContext to Accept JsonObject for attributes

gaurav-rana:
In the `ConditionEvaluator.evaluateCondition` method, we currently convert `attributesJson` and `conditionJson` from `String` to `JsonObject`. However, since this method is also recursive, these conversions are performed multiple times during a single feature evaluation. This PR addresses this issue by moving `attributesJson` and `conditionJson` to `JsonObject` earlier to avoid repeated conversions.

Additionally, this PR updates the `GBContext` constructor to accept `userAttributes` as a `JsonObject`. Currently, the constructor expects userAttributes as a `String`, which is inefficient for applications that frequently update userAttributes, as it requires converting the `JsonObject` to a `String` each time. Accepting userAttributes as a `JsonObject` improves efficiency and aligns with the implementation in other SDKs (e.g., [Go](https://github.com/growthbook/growthbook-golang/blob/main/types.go#L7)).

### DIFF
--- a/lib/src/main/java/growthbook/sdk/java/IConditionEvaluator.java
+++ b/lib/src/main/java/growthbook/sdk/java/IConditionEvaluator.java
@@ -1,5 +1,7 @@
 package growthbook.sdk.java;
 
+import com.google.gson.JsonObject;
+
 interface IConditionEvaluator {
-    Boolean evaluateCondition(String attributesJsonString, String conditionJsonString);
+    Boolean evaluateCondition(JsonObject attributesJson, JsonObject conditionJson);
 }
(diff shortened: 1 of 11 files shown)


lib/src/main/java/growthbook/sdk/java/IConditionEvaluator.java
```diff
@@ -1,5 +1,7 @@
 package growthbook.sdk.java;
 
+import com.google.gson.JsonObject;
+
 interface IConditionEvaluator {
-    Boolean evaluateCondition(String attributesJsonString, String conditionJsonString);
+    Boolean evaluateCondition(JsonObject attributesJson, JsonObject conditionJson);
```
Bohdan-Kim, line 6: So, the signature of method with `default` visibility was changed

Bohdan-Kim (approved):
These changes look good for me

gaurav-rana:
Great! let me know when we would be able to merge the PR.

Bohdan-Kim:
It seems like we can merge it now.

PR #97, opened 2024-08-06
Release v0.9.91

vazarkevych:


gvillo:
it would be awesome if this can be releases ASAP, I got 2GB of logs just for yesterday because of this bug while releasing this SDK

gvillo (approved):

msdousti24:
I also would love for this to be merged. We had over 700K errors in our past week logs, and they were only caused by GrowthBook.

PR #102, opened 2024-10-10
Patch v0.9.92

vazarkevych:


Bohdan-Kim (approved):

Bohdan-Kim (approved):

lib/src/main/java/growthbook/sdk/java/Constants.java
```diff
@@ -0,0 +1,15 @@
...
+    String ACCEPT = "Accept";
+    String SSE_HEADER = "text/event-stream";
+    String ENABLED = "enabled";
+    String ENCRYPTED_FEATURES_KEY = "encryptedFeatures";
+    String ENCRYPTED_SAVED_GROUPS_KEY = "encryptedSavedGroups";
+    String FEATURE_KEY = "features";
+    String SAVED_GROUP_KEY = "savedGroups";
+    String DATA_SSE_KEY = "data:";
+    Integer QUANTITY_TO_CUT_SSE = 5;
+}
```
madhuchavva, line 15: please move these constant definitions to the respective files and not create a dedicated interface for constants. Makes it easy to maintain.

lib/src/main/java/growthbook/sdk/java/GbConstants.java
```diff
@@ -0,0 +1,6 @@
+package growthbook.sdk.java;
+
+interface GbConstants {
+    String ENABLED = "enabled";
+    Integer QUANTITY_TO_CUT_SSE = 5;
+}
```
madhuchavva, line 6: Can we move the `QUANTITY_TO_CUT_SSE` to repository and ENABLED to transformer util class?

vazarkevych: Hi, @madhuchavva. We added changes, could you please take a look on that?

madhuchavva (approved):

PR #105, opened 2024-11-13
Manual forced feature

vazarkevych:
add manual forced feature to evaluation process


madhuchavva:
@vazarkevych Can we discuss about these global forcedFeatures?

vazarkevych:
Hi, @madhuchavva . Yes, let's setup a call for that. I will write to you.

PR #106, opened 2024-11-20
Changes to enable Multi Context Support

madhuchavva:
Changes to support GrowthBookMultiUser

### GrowthBookMultiUser
GrowthBookMultiUser lets you share the same instance for all requests with an ability to accept the user attributes 
while calling the feature methods like `isOn()`. This GrowthBookMultiUser instance is decoupled from the `GBContext`, 
creates a singleton featureRepository based on your refreshStrategy and uses the latest features at the time of
evaluation, all managed internally.

```java

// build options to configure your Growthbook instance
Options options = Options.builder()
        .apiHost("https://cdn.growthbook.io")
        .clientKey("sdk-abc123")
        .build();

// Create growthbook instance using the options you need
GrowthBookMultiUser gb = new GrowthBookMultiUser(options);

// call the init method to load features 
gb.initialize();

gb.isOn("featureKey", UserContext.builder()
    .attributesJson("{\"id\" : \"req.user.id\"}").build()
);
```

Compare this to how it's done today!

```java

GBFeaturesRepository featuresRepository = GBFeaturesRepository
    .builder()
    .apiHost("https://cdn.growthbook.io")
    .clientKey("<environment_key>") // replace with your client key
    .encryptionKey("<client-key-for-decrypting>") // optional, nullable
    .refreshStrategy(FeatureRefreshStrategy.SERVER_SENT_EVENTS) // optional; options: STALE_WHILE_REVALIDATE, SERVER_SENT_EVENTS (default: STALE_WHILE_REVALIDATE)
    .build();

// Optional callback for getting updates when features are refreshed
featuresRepository.onFeaturesRefresh(new FeatureRefreshCallback() {
    @Override
    public void onRefresh(String featuresJson) {
        System.out.println("Features have been refreshed");
        System.out.println(featuresJson);
    }
});

try {
    featuresRepository.initialize();
} catch (FeatureFetchException e) {
    // TODO: handle the exception
    e.printStackTrace();
}

// Initialize the GrowthBook SDK with the GBContext and features
GBContext context = GBContext
    .builder()
    .featuresJson(featuresRepository.getFeaturesJson())
    .attributesJson(userAttributesJson)
    .build();

GrowthBook growthBook = new GrowthBook(context);

growthBook.isOn("featureKey");
```


lib/src/main/java/growthbook/sdk/java/multiusermode/GrowthBookMultiUser.java
```diff
@@ -0,0 +1,131 @@
...
+    }
+
+    private GlobalContext getGlobalContext() {
+        GlobalContext globalContext =  GlobalContext.builder()
+            .featuresJson(repository.getFeaturesJson())
+            .build();
+
+        globalContext.loadEncryptedFeatures(this.options.getDecryptionKey());
+        return globalContext;
+    }
```
jdorn, line 78: For performance, I think we will want to only have a single `GlobalContext` object created during initialization and re-used for every evaluation.

lib/src/main/java/growthbook/sdk/java/multiusermode/configurations/GlobalContext.java
```diff
@@ -0,0 +1,48 @@
...
+            // TBD:M experiments
+        }
+        return this.features;
+    }
+
+    public void loadEncryptedFeatures(String decryptionKey) {
+        if (this.features == null) {
+            this.features = TransformationUtil.transformEncryptedFeatures(this.featuresJson, decryptionKey);
+        }
+    }
```
jdorn, line 47: Getting / decrypting features seems like something that should be done during init, not lazily loaded.  If we did that, we wouldn't need to store `featuresJson` here at all, just the final `features` object.

lib/src/main/java/growthbook/sdk/java/multiusermode/configurations/Options.java
```diff
@@ -0,0 +1,141 @@
...
+                   @Nullable String decryptionKey,
+                   @Nullable List<String> stickyBucketIdentifierAttributes,
+                   @Nullable StickyBucketService stickyBucketService,
+                   @Nullable TrackingCallbackWithUser trackingCallBackWithUser,
+                   @Nullable FeatureUsageCallbackWithUser featureUsageCallbackWithUser,
+                   @Nullable FeatureRefreshStrategy refreshStrategy,
+                   @Nullable FeatureRefreshCallback featureRefreshCallback) {
+        this.enabled = enabled == null || enabled;
+        this.isQaMode = isQaMode != null && isQaMode;
+        this.isCacheDisabled = isCacheDisabled == null || isCacheDisabled;
```
jdorn, line 37: This looks wrong to me.  it's defaulting cache to be disabled when `isCacheDisabled` is null.  I would think we would want it enabled by default instead.

lib/src/main/java/growthbook/sdk/java/multiusermode/configurations/OptionsDup.java
```diff
@@ -0,0 +1,37 @@
...
+import lombok.extern.slf4j.Slf4j;
+
+import javax.annotation.Nullable;
+
+/**
+ * Options you use to configure your growthbook instance
+ */
+@Data
+@Slf4j
+public class OptionsDup {
```
jdorn, line 14: What is this class used for?

lib/src/main/java/growthbook/sdk/java/multiusermode/configurations/UserContext.java
```diff
@@ -0,0 +1,53 @@
...
+        if (this.attributes == null) {
+            return new JsonObject();
+        }
+
+        return this.attributes;
+    }
+
+    public Map<String, Integer> getForcedVariationsMap() {
+        return this.forcedVariationsMap == null ? new HashMap<>() : this.forcedVariationsMap;
+    }
```
jdorn, line 47: Most of the time, `forcedVariationsMap` is going to be null.  This will create a new `HashMap` every time an experiment is evaluated.  For performance, should we do this once instead and re-use the object?  Maybe creating an empty HashMap is super lightweight and it doesn't matter, so feel free to ignore if that's the case.

lib/src/main/java/growthbook/sdk/java/GBFeaturesRepository.java
```diff
@@ -44,7 +44,7 @@ public class GBFeaturesRepository implements IGBFeaturesRepository {
 
     @Nullable
     @Getter
-    private final String encryptionKey;
+    private final String decryptionKey;
```
jdorn, line 47: I agree with this change, but it will break any existing implementation that is using the old naming.  I think we should have both `decryptionKey` and `encryptionKey` getters/setters work and mark the old encryptionKey ones as deprecated.  Then, in a future version we can remove them.

lib/src/main/java/growthbook/sdk/java/ExperimentEvaluator.java
```diff
@@ -126,49 +124,57 @@ public <ValueType> ExperimentResult<ValueType> evaluateExperiment(Experiment<Val
...
             }
 
             // 10. Exclude if prerequisites are not met
             List<ParentCondition> parenConditions = experiment.getParentConditions();
             if (parenConditions != null) {
                 for (ParentCondition parentCondition : parenConditions) {
+
+                    UserContext user = UserContext.builder()
+                            .attributes(jsonUtils.gson.fromJson(parentCondition.getCondition(), JsonObject.class))
+                            .build();
```
jdorn, line 160: This doesn't feel right to me.  I'm pretty sure we don't want to modify the user's attributes or create a subContext at all here.  It looks like the original implementation was wrong here as well.

lib/src/main/java/growthbook/sdk/java/ExperimentEvaluator.java
```diff
@@ -1,36 +1,35 @@
...
 /**
  * <b>INTERNAL</b>: Implementation of experiment evaluation
  */
 @Slf4j
-class ExperimentEvaluator implements IExperimentEvaluator {
+public class ExperimentEvaluator implements IExperimentEvaluator {
 
     private final ConditionEvaluator conditionEvaluator = new ConditionEvaluator();
     private final GrowthBookJsonUtils jsonUtils = GrowthBookJsonUtils.getInstance();
+    private final Set<String> trackedExperiments = new HashSet<>();
```
jdorn, line 19: I'm a little worried about having this here.  With the MultiUser mode, this set will continue to grow unbounded.  Millions of users = millions of entries in the set.

In JavaScript, I solved this by removing this logic completely and leaving it up to the user to de-duplicate tracking calls.  Another option is to turn this into an LRU cache that only keeps the most recent X entries.

Note: This was not an issue before since we were persisting this in the GrowthBook instance, which was for a single user and short lived.

lib/src/main/java/growthbook/sdk/java/FeatureEvaluator.java
```diff
@@ -374,13 +382,14 @@ public <ValueType> FeatureResult<ValueType> evaluateFeature(
...
-            if (featureUsageCallback != null) {
-                featureUsageCallback.onFeatureUsage(key, defaultValueFeatureResult);
+            if (featureUsageCallbackWithUser != null) {
+                featureUsageCallbackWithUser.onFeatureUsage(key, defaultValueFeatureResult, context.getUser());
             }
 
             // Return (value = defaultValue or null, source = defaultValue)
             return defaultValueFeatureResult;
         } catch (Exception e) {
+            e.printStackTrace();
```
jdorn, line 392: Is this an intentional change?

lib/src/main/java/growthbook/sdk/java/GBFeaturesRepository.java
madhuchavva: Good catch, it would be a breaking change. I'll fix it.

lib/src/main/java/growthbook/sdk/java/multiusermode/configurations/UserContext.java
madhuchavva: creating an empty hashmap is far more simple and super lighweight. it would be gc'ed as soon as the object holding it dies.

lib/src/main/java/growthbook/sdk/java/multiusermode/GrowthBookMultiUser.java
madhuchavva: `GlobalContext` is not a heavy object and apart from unmarshalling the features json, there are no heavy operations. If we change the implementation to re-use the same instance over n over, we'll have to create hooks to read the latest features from the repository to override the features. And this operation would have to synchronized with any evaluations that are in-progress. Otherwise we will run into issues using this in the multi threaded context.

lib/src/main/java/growthbook/sdk/java/GrowthBookUtils.java
```diff
@@ -446,24 +447,19 @@ public static Boolean inRange(Float n, BucketRange range) {
      * </ul>
      *
      * @param filters            List<Filters></Filters>
-     * @param attributeOverrides JsonObject
-     * @param context            GBContext
+     * @param attributes         JsonObject
      * @return check if user filtered
      */
-    public static Boolean isFilteredOut(List<Filter> filters, JsonObject attributeOverrides, GBContext context) {
+    public static Boolean isFilteredOut(List<Filter> filters, JsonObject attributes) {
```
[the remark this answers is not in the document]
madhuchavva: I didn't find any documentation for it and it differs from the current public methods signatures. So I thought It'd be a good idea to fix it.

lib/src/main/java/growthbook/sdk/java/GrowthBook.java
```diff
@@ -72,12 +83,44 @@ public GrowthBook() {
...
+
+        // build user context
+        UserContext userContext = UserContext.builder()
+                .attributes(this.context.getAttributes())
+                .stickyBucketAssignmentDocs(this.context.getStickyBucketAssignmentDocs())
+                .forcedVariationsMap(this.context.getForcedVariationsMap())
+                .build();
+
+        this.evaluationContext = new EvaluationContext(globalContext, userContext,
+                new EvaluationContext.StackContext(), options);
```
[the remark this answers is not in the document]
madhuchavva: This creates a new StackContext instance for every evaluation. we're not reusing it. It's just a static declaration in the EvaluationContext as it is an inner class.

lib/src/main/java/growthbook/sdk/java/FeatureEvaluator.java
madhuchavva: yeah, a debug statement. I'll remove it!

lib/src/main/java/growthbook/sdk/java/ExperimentEvaluator.java
madhuchavva: I'm happy to convert that into an LRU like solution. I wasn't completely clear about the actual need behind this though! Today, in the current implementation adds all/any experiment to this set and doesn't clear it.

lib/src/main/java/growthbook/sdk/java/multiusermode/configurations/GlobalContext.java
madhuchavva: yeah, I agree! but the lombok library isn't supporting custom methods. I'll have to write custom builders to achieve this. Let me try.

lib/src/main/java/growthbook/sdk/java/multiusermode/GrowthBookMultiUser.java
jdorn: The unmarshalling might be expensive for large payloads, right? 

The payload changes very infrequently (max once a minute, but usually more like hours or days), but feature evaluations might happen thousands of times per second or more.

Seems like we should only do the expensive unmarshalling part when the payload changes and not for every eval.

For background, one of the main reasons we're doing this refactor is for customers with massive request volume running into performance bottlenecks. So anything we can do to reduce work for each eval is important.

madhuchavva: Yeah, I found a way to optimize this in a simple way. Almost done with the changes, I'll update you!

lib/src/main/java/growthbook/sdk/java/GrowthBook.java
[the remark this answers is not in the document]
jdorn: Aren't we re-using `this.evaluationContext` multiple times though?  If you evaluate 5 feature flags in a row, it will use the same context for all 5.  The stack context needs to be reset for every evaluation, otherwise it could cause conflicts.

lib/src/main/java/growthbook/sdk/java/ExperimentEvaluator.java
jdorn: But in the current implementation, the GrowthBook instances are short-lived (one per-request), so the set will not grow unbounded.  We're changing that with the multi-user class, so will need to implement our own method of keeping the set from growing too large.

lib/src/main/java/growthbook/sdk/java/GrowthBook.java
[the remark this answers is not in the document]
madhuchavva: Ohh.. missed this file in the commit. just a sec.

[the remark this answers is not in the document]
madhuchavva: This is addressed.

lib/src/main/java/growthbook/sdk/java/GrowthBook.java
```diff
@@ -117,10 +117,16 @@ private void initializeEvalContext() {
                 new EvaluationContext.StackContext(), options);
     }
 
+    private EvaluationContext getEvaluationContext() {
+        // Reset the stackContext for every evaluation.
+        this.evaluationContext.setStack(new EvaluationContext.StackContext());
```
jdorn, line 122: I don't know how Java handles concurrency, but will this cause any race conditions if 2 evaluations are running at the same time?  Or is it always single threaded and not an issue?

madhuchavva: Our present `GrowthBook` Instance is not built for concurrency. If the same instance is used by multiple threads a lot of changes are needed to address the data inconsistencies and race conditions. To support the concurrent execution, we'll have to synchronize the execution and re-org the way the current evaluation is happening.

madhuchavva: @jdorn Merging the PR, based on our convo.

PR #108, opened 2025-01-09
Manual forced feature 

vazarkevych:
add manual forced feature to evaluation process

### DIFF
--- a/lib/src/main/java/growthbook/sdk/java/multiusermode/GrowthBookClient.java
+++ b/lib/src/main/java/growthbook/sdk/java/multiusermode/GrowthBookClient.java
@@ -9,6 +9,11 @@ import growthbook.sdk.java.multiusermode.configurations.UserContext;
 import growthbook.sdk.java.multiusermode.util.TransformationUtil;
 import lombok.extern.slf4j.Slf4j;
 
+import java.util.ArrayList;
+import java.util.HashMap;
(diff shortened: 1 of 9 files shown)


madhuchavva (approved):

lib/src/main/java/growthbook/sdk/java/multiusermode/GrowthBookClient.java
```diff
@@ -32,9 +32,10 @@ public GrowthBookClient() {
     public GrowthBookClient(Options opts) {
         this.options = opts == null ? Options.builder().build() : opts;
 
+        this.assigned = new HashMap<>();
```
madhuchavva, line 35: what are assigned experiments and where are we using this assigned?? Can you some details on how the definition for "AssignedExperiment" looks like?

vazarkevych: AssignedExperiment is a POJO class that previously contained the experiment and experiment result. We created this class specifically for the subscriptions logic described in the [documentation](https://docs.growthbook.io/lib/build-your-own#subscriptions). After your [suggestion](https://github.com/growthbook/growthbook-sdk-java/pull/109#discussion_r1918581843), this class now contains only three variables: experimentKey, inExperiment, and variationId.

PR #109, opened 2025-01-10
Improvements for Remote Evaluation 

vazarkevych:
Fix Trackdata in evaluation process add remote eval feature, fix tests add test for LocalGbFeatureRepository and NativeJavaGbFeatureRepository; add ability to pass requestBody for new variant of GBClient; fix test, add remote eval feature to new flow of initialize GB


lib/src/main/java/growthbook/sdk/java/multiusermode/GrowthBookClient.java
```diff
@@ -141,12 +153,30 @@ public <ValueType> ExperimentResult<ValueType> run(Experiment<ValueType> experim
...
     }
 
     public void subscribe(ExperimentRunCallback callback) {
         this.callbacks.add(callback);
     }
+
+    private <ValueType> void fireSubscriptions(Experiment<ValueType> experiment, ExperimentResult<ValueType> result) {
+        String key = experiment.getKey();
+        // If assigned variation has changed, fire subscriptions
+        AssignedExperiment prev = this.assigned.get(key);
```
madhuchavva, line 168: Instead of storing the entire experiment and experimentResult, we should come up with a simple/light weight implementation to uniquely identify the experiments tracked. I can provide sample code if needed.

lib/src/main/java/growthbook/sdk/java/GBFeaturesRepository.java
```diff
@@ -190,11 +204,13 @@ public GBFeaturesRepository(
         // Build the endpoints from the apiHost and clientKey
         this.featuresEndpoint = apiHost + "/api/features/" + clientKey;
         this.eventsEndpoint = apiHost + "/sub/" + clientKey;
+        this.remoteEvalEndPoint = apiHost + "/api/eval/" + clientKey;
 
         this.encryptionKey = decryptionKey;
         this.decryptionKey = decryptionKey;
 
         this.swrTtlSeconds = swrTtlSeconds == null ? 60 : swrTtlSeconds;
+        this.requestBodyForRemoteEval = requestBodyForRemoteEval != null ? requestBodyForRemoteEval : new HashMap<>();
```
madhuchavva, line 213: Is there a structure we can use for it instead of having a hashmap?

lib/src/main/java/growthbook/sdk/java/multiusermode/GrowthBookClient.java
```diff
@@ -30,6 +40,7 @@ public GrowthBookClient(Options opts) {
         this.featureEvaluator = new FeatureEvaluator();
         this.experimentEvaluatorEvaluator = new ExperimentEvaluator();
         this.callbacks = new ArrayList<>();
+        this.requestBodyForRemoteEval = new HashMap<>();
```
madhuchavva, line 43: What's the structure for it? Is there a way we can incorporate it in Options and still follow the initialization steps of using configurable options to create a client?

lib/src/main/java/growthbook/sdk/java/multiusermode/GrowthBookClient.java
vazarkevych: We removed experiment and result from the AssignedExperiment class, and now store only the variables needed for these functions: experimentKey, inExperiment, and variationId

lib/src/main/java/growthbook/sdk/java/GBFeaturesRepository.java
vazarkevych: We replaced the HashMap with a Payload class

lib/src/main/java/growthbook/sdk/java/multiusermode/GrowthBookClient.java
vazarkevych: We replaced the HashMap with a Payload class, which will be configured with data from Options and passed to the GBFeatureRepository.
This data is used for a POST request in the remote evaluation feature. An example of this payload looks like this:
```
{
    "attributes"?: {"attrKey":"attrValue"},
    "forcedFeatures"?: [["test3", 1], ["my-feature1", 2]],
    "forcedVariations"?: {"gbdemo-checkout-layout":0},
    "url"?: "https://ray-guided-duly.ngrok-free.app/"
}
```

lib/src/main/java/growthbook/sdk/java/multiusermode/configurations/Options.java
```diff
@@ -46,6 +54,9 @@ public Options(@Nullable Boolean enabled,
         this.featureUsageCallbackWithUser = featureUsageCallbackWithUser;
         this.refreshStrategy = refreshStrategy;
         this.featureRefreshCallback = featureRefreshCallback;
+        this.attributes = attributes;
+        this.forcedFeatureValues = forcedFeatureValues;
+        this.forcedVariationsMap = forcedVariationsMap;
```
madhuchavva, line 59: Attributes, forcedFeatureValues and variations are intentionally separated out into UserContext, such that we offer a flexibility to change it for every evaluation if need be whenever users invoke eval methods.

`Options` should hold parameters that are used to configure the GrowthBook and GrowthBookClient instances and should not be mixed up with UserContext.

lib/src/main/java/growthbook/sdk/java/Payload.java
```diff
@@ -0,0 +1,26 @@
...
+@AllArgsConstructor
+public class Payload {
+    @Nullable
+    private JsonObject attributes;
+    @Nullable
+    private List<List<Object>> forcedFeatures;
+    @Nullable
+    private Map<String, Integer> forcedVariations;
+    @Nullable
+    private String url;
```
madhuchavva, line 25: how is this data different from UserContext? Can this be different for every remote eval? If so, why should it be supplied or configured, at the time of client initialization?

lib/src/main/java/growthbook/sdk/java/GBFeaturesRepository.java
```diff
@@ -289,6 +328,10 @@ public void initialize(Boolean retryOnFailure) throws FeatureFetchException {
                 fetchFeatures();
                 initializeSSE(retryOnFailure);
                 break;
+
+            case REMOTE_EVAL_STRATEGY:
+                fetchForRemoteEval(this.payload);
```
madhuchavva, line 333: Can this be implemented in a different way? why do we need to pre-fetch info for remoteEval?

madhuchavva (changes requested):

lib/src/main/java/growthbook/sdk/java/GBFeaturesRepository.java
vazarkevych: We believe this cannot be implemented in a different way. Remote evaluation requires a distinct configuration approach, as follows:

- A proxied apiHost needs to be provided.
- A POST request must be used instead of GET.
- A request body is essential because, without passing attributes in the request body, fetching evaluated features is not possible.
- Additionally, URL building for remote evaluation follows a different approach.

To enhance flexibility, we suggest adding a setters: for _globalAttributes_, _globalForceVariations_, _globalForceFeatures_ - and _refresh function_ for remote evaluation. This would allow the required parameters to be dynamically passed in the request body. Such a feature could also prove useful for GET requests that support a refreshing mechanism.

lib/src/main/java/growthbook/sdk/java/multiusermode/configurations/Options.java
vazarkevych: We can see your point, but in the TypeScript class, Options also include: forcedVariations forcedFeatureValues attributes. We assume it is designed this way because, in cases where a feature needs to be evaluated remotely, these variables must be present in the configuration. Additionally, we noticed that in TypeScript, the attributes from UserContext are merged with the attributes from Options when creating the EvaluationContext.

lib/src/main/java/growthbook/sdk/java/Payload.java
vazarkevych: This data is required for configuration and to successfully fetch features during remote evaluations.

If there is a need to modify this data, we provide setters such as setGlobalAttributes, setGlobalForceFeatures, and setGlobalForceVariations. After updating the data, you can invoke the refreshForRemoteEval function to pass the updated information to the request body.

madhuchavva (approved):

lib/src/main/java/growthbook/sdk/java/GBFeaturesRepository.java
madhuchavva: we'll have it optimized in coming releases. For now, I'm ok with this approach.

lib/src/main/java/growthbook/sdk/java/Payload.java
madhuchavva: we'll have it optimized in coming releases. For now, I'm ok with this approach.

PR #112, opened 2025-01-10
Fix evaluating force property in rule when rule is JsonObject and defined as null

vazarkevych:
Fix evaluating force property in rule when rule is JsonObject and defined as null

### DIFF
--- a/lib/src/main/java/growthbook/sdk/java/FeatureEvaluator.java
+++ b/lib/src/main/java/growthbook/sdk/java/FeatureEvaluator.java
@@ -141,7 +141,9 @@ public class FeatureEvaluator implements IFeatureEvaluator {
             }
 
             // Loop through the feature rules (if any)
-            for (FeatureRule<ValueType> rule : feature.getRules()) {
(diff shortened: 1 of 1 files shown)


lib/src/main/java/growthbook/sdk/java/FeatureEvaluator.java
```diff
@@ -431,4 +433,8 @@ private void addFeatureToEvalStack(String featureKey, EvaluationContext context)
         context.getStack().setId(featureKey);
         context.getStack().getEvaluatedFeatures().add(featureKey);
     }
+
+    private static boolean hasForceProperty(JsonElement featureJson, int i) {
+        return featureJson.getAsJsonObject().get("rules").getAsJsonArray().get(i).getAsJsonObject().has("force");
```
madhuchavva, line 438: Can we fix the marshalling to make the FeatureRule read from this place, so that attribute "force" is properly mapped in FeatureRule ?

vazarkevych: Yes, sure. We are now working on that.

PR #115, opened 2025-01-24
Throw exception on HTTP errors

martinth:
Adds a new FeatureFetchErrorCode called HTTP_RESPONSE_ERROR that is used on non-200 HTTP status codes and also logs the server response. One of the common instances this happens is when you use a wrong SDK key (which I managed to do by copy and pasting to few characters 😬).

They were previously mangled into FeatureFetchErrorCode.CONFIGURATION_ERROR with a very non-descriptive message of "No features found". And since you couldn't see the API response anywhere you were essentially blind on the "why" unless you were able to place a breakpoint into the `onResponseJson`.

Logging the response might be something to argue about but I think it will really help with finding the root cause.

---

This is my first contribution here so let me know if you want anything changed. We can also take this PR as a chance to discuss if this even a good idea.

### DIFF
--- a/lib/src/main/java/growthbook/sdk/java/FeatureFetchException.java
+++ b/lib/src/main/java/growthbook/sdk/java/FeatureFetchException.java
@@ -34,6 +34,11 @@ public class FeatureFetchException extends Exception {
         CONFIGURATION_ERROR,
 
         /**
+         * - the server responded with an HTTP error
+         */
+        HTTP_RESPONSE_ERROR,
+
+        /**
          * - there was no response body
          */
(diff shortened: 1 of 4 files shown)


vazarkevych:
Hello, @martinth!

Thank you for your contribution! We completely agree that providing better error handling for non-200 HTTP responses will be helpful for users to diagnose issues, especially when an incorrect SDK key is used. Your PR makes this process much clearer by introducing a dedicated error code and logging the server response.

We appreciate your effort, and we'll include your changes in a future release. If we need any additional adjustments, we’ll be sure to let you know.

PR #135, opened 2025-04-03
fix: tracking callback invocation

vazarkevych:
## Summary
  
  Fixes duplicate invocation of the tracking callback (`TrackingCallback.onTrack`)
  during **local** experiment evaluation. Previously the callback fired on *every*
  evaluation of the same assignment instead of exactly once.

  ## Problem
  
  `ExperimentEvaluator.isExperimentTracked(...)` recorded each experiment in the
  tracker but **always returned `false`**. Because the call site fires the callback
  when `!isExperimentTracked(...)`, `onTrack` re-fired on every evaluation for the
  same `(hashAttribute, hashValue, experimentKey, variationId)` combination.                                                                                                                           

  ## Fix

  - `isExperimentTracked(...)` now returns whether the assignment was already
    tracked (delegates to the existing `alreadyTracked(...)` helper), so `onTrack`
    fires exactly once per unique assignment.
  - De-duplication is handled solely by the bounded `ExperimentTracker`
    (Guava LRU cache, `maximumSize = 30`).


lib/src/main/java/growthbook/sdk/java/evaluators/ExperimentEvaluator.java
```diff
@@ -394,23 +391,6 @@ private <ValueType> ExperimentResult<ValueType> getExperimentResult(
...
-        String experimentKey = experiment.getKey();
-
-        String key = (
-                result.getHashAttribute() != null ? result.getHashAttribute() : "")
-                + (result.getHashValue() != null ? result.getHashValue() : "")
-                + (experimentKey + result.getVariationId());
-
-        // Add the experiment to the tracker if it doesn't exist.
-        if (!experimentTracker.isExperimentTracked(key)) {
-            experimentTracker.trackExperiment(key);
```
madhuchavva, line 408: We use an Experiment Tracker to keep track of the most recent experiments set with a bounded max val of 30(which can be changed). Changing this implementation to a unbounded hashSet will grow significantly when there are multiple instances of GrowthBookClient is created.

lib/src/main/java/growthbook/sdk/java/multiusermode/configurations/Options.java
```diff
@@ -157,6 +157,7 @@ public Options(@Nullable Boolean enabled,
     @Nullable
     private Map<String, Object> globalForcedFeatureValues;
 
+    private ExperimentHelper experimentHelper = new ExperimentHelper();
```
madhuchavva, line 160: `Options` is not a place for initializing experimentHelper. Options technically provides all configurable parameters that influence the behavior of GrowthBook instance.

madhuchavva (changes requested):

lib/src/main/java/growthbook/sdk/java/multiusermode/configurations/Options.java
vazarkevych: Ok, we have moved the Experiment Tracker to the Global Context. We need this instance to be a singleton in order to track experiments and their results properly — only once. Previously, it was being recreated during every evaluation process

lib/src/main/java/growthbook/sdk/java/evaluators/ExperimentEvaluator.java
vazarkevych: We’ve restored your implementation of the Experiment Tracker as suggested.

PR #162, opened 2025-05-26
Updating features fix

vazarkevych:
The issue was that feature flags were never refreshed after the initial load. I fixed this by triggering a feature refresh whenever the evaluation context is set or updated.

This follows a stale-while-revalidate approach — it uses the cached features immediately, while asynchronously fetching and applying the latest ones in the background.


madhuchavva:
This issue is addressed in the PR #168

PR #163, opened 2025-06-13
fix updating features without cache

vazarkevych:
This PR ensures that feature updates continue to work even when `.isCacheDisabled(true)` is set. Previously, disabling the cache also stopped background refreshes. 

Also, a **cyclical call** to `repository.getParsedSavedGroups()` inside `onRefresh()` was removed to prevent unnecessary repeated calls and improve clarity.


madhuchavva:
Fixes are covered in #168

PR #167, opened 2025-08-18
feat: improve sticky bucketing

vazarkevych:
In this pr we:
- Added FileStickyBucketServiceImpl that implements StickyBucketService using GbCacheManager to persist all sticky bucket assignments in a single file.
- Supports asynchronous fetching and saving via CompletableFuture.

### DIFF
--- a/lib/src/main/java/growthbook/sdk/java/GrowthBook.java
+++ b/lib/src/main/java/growthbook/sdk/java/GrowthBook.java
@@ -5,7 +5,9 @@ import java.util.HashMap;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Objects;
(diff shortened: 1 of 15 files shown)


madhuchavva (changes requested):
@vazarkevych Thoughtful contribution on providing File-based storage. But it involves a lot of risks that are needed to be taken care such as Concurrency bottlenecks (read-modify-write), Performance issues (full file read/write per operation) and Corruption risks (if the process crashes midway, the data is lost and the file isn't readable anymore).

On top of it, This PR replaces the current synchronous behavior for `StickyBucketingService` interface with async methods, which breaks the existing/custom implementations and concurrent access to a shared file needs synchronization. 

Can you provide more details about this feature request?

PR #206, opened 2026-04-22
Add plugins support for both legacy and new clients

madhuchavva:
## Summary

Adds support for GrowthBook plugins and ports the built-in tracking plugin pattern for the legacy client.

## What changed

- Wired plugin lifecycle into both single-user (`GrowthBook` / `GBContext`) and multi-user (`GrowthBookClient` / `Options`) modes
- Routed experiment and feature evaluation events through the shared evaluator path so plugins receive the same events in both modes
- Preserved existing callback behavior by chaining plugin dispatch after `TrackingCallback` / `FeatureUsageCallback`
- Added a built-in `GrowthBookTrackingPlugin` that batches `experiment_viewed` and `feature_evaluated` events and POSTs them to the ingest endpoint
- Added tests for plugin lifecycle, evaluator integration, batching, timer-based flushing, close-time flushing, and failure/no-op behavior

## Wire contract

- Endpoint: `POST {ingestorHost}/events`
- Default host: `https://us1.gb-ingest.com`
- Body: `{ "client_key": "...", "events": [...] }`
- Headers: `Content-Type: application/json`, `User-Agent: growthbook-java-sdk/{version}`
- Event types: `experiment_viewed`, `feature_evaluated`
- Batch defaults: `batchSize=100`, `batchTimeout=10s`
- `close()` performs a synchronous final flush
- Initialization failures degrade to no-op behavior


lib/src/main/java/growthbook/sdk/java/evaluators/ExperimentEvaluator.java
```diff
@@ -312,6 +313,11 @@ public <ValueType> ExperimentResult<ValueType> evaluateExperiment(Experiment<Val
             if (trackingCallBackWithUser != null) {
                 trackingCallBackWithUser.onTrack(experiment, result, context.getUser());
             }
+
+            PluginRegistry pluginRegistry = context.getOptions().getPluginRegistry();
```
vazarkevych: `evaluateExperiment` is already one of the largest methods in the codebase and
  carries a lot of line-by-line "what" comments that restate the code. This change
  adds another inline block to it, which pushes it further in the wrong direction.

  Two things would keep it in check:

  1. Extract the dispatch instead of inlining it — mirror the `dispatchFeatureUsage(...)`
     helper this same PR already introduced in `FeatureEvaluator`, e.g.
     `dispatchExperimentViewed(context, experiment, result)`. The method body then
     gains one call, not a new null-check block, and both evaluators stay consistent.

  2. While touching this block, trim the redundant comments (the ones that just
     re-describe the key formula / the callback) rather than growing the method. The
     goal here should be fewer lines and less noise, not more.

  Net: the plugin hook is welcome, but let's add it via a small helper and take the
  opportunity to shrink this method rather than extend it.

lib/src/main/java/growthbook/sdk/java/evaluators/FeatureEvaluator.java
```diff
@@ -458,6 +441,21 @@ private void addFeatureToEvalStack(String featureKey, EvaluationContext context)
         context.getStack().getEvaluatedFeatures().add(featureKey);
     }
 
+    private <ValueType> void dispatchFeatureUsage(
```
vazarkevych, line 403: `ValueType` isn't idiomatic Java. The convention for type parameters is a
  single uppercase letter (`T`, or `V` for a value), not a descriptive word — the JLS
  naming guideline exists precisely so type variables are visually distinct from
  class names.

lib/src/main/java/growthbook/sdk/java/evaluators/FeatureEvaluator.java
```diff
@@ -117,9 +110,7 @@ public <ValueType> FeatureResult<ValueType> evaluateFeature(
             // Unknown key, return empty feature
```
vazarkevych, line 109: Since this PR is already refactoring `evaluateFeature` (swapping the inline
  callback calls for `dispatchFeatureUsage`), it's a good moment to drop the
  redundant comments in the method while you're in there. Most of them just restate
  the next line ("Unknown key, return empty feature", "When key exists but there is
  no value…", etc.) and add noise to an already long method.

lib/src/main/java/growthbook/sdk/java/evaluators/FeatureEvaluator.java
```diff
@@ -284,16 +263,26 @@ public <ValueType> FeatureResult<ValueType> evaluateFeature(
                     // Call the tracking callback with all the track data
                     List<TrackData<ValueType>> trackData = rule.getTracks();
                     TrackingCallbackWithUser trackingCallBackWithUser = context.getOptions().getTrackingCallBackWithUser();
+                    PluginRegistry pluginRegistry = context.getOptions().getPluginRegistry();
```
vazarkevych: Same issue again — this just bolts more inline logic onto an already large evaluateFeature. Would you mind extracting the onTrack + fireExperimentViewed dispatch into a shared helper instead of growing the method further? Should keep things a bit tidier going forward, thanks!

lib/src/main/java/growthbook/sdk/java/evaluators/FeatureEvaluator.java
```diff
@@ -284,16 +263,26 @@ public <ValueType> FeatureResult<ValueType> evaluateFeature(
...
 
                     // If this was a remotely evaluated experiment, fire the tracking callbacks
-                    if (trackData != null && trackingCallBackWithUser != null) {
-                        trackData.forEach(t ->
+                    if (trackData != null) {
+                        trackData.forEach(t -> {
```
vazarkevych: Since main already solved this in #216 with experimentEvaluator.fireRemoteEvaluationTracks(...) (dedupes exposures, null-guards payloads), it'd be worth rebasing onto that instead of adding a parallel helper here. Could you route the plugin's fireExperimentViewed through that existing path once you've rebased? Should save us from maintaining two versions of the same logic.

lib/src/main/java/growthbook/sdk/java/plugin/tracking/GrowthBookTrackingPlugin.java
```diff
@@ -0,0 +1,305 @@
...
+    private static ThreadFactory daemonFactory(String prefix) {
+        return r -> {
+            Thread t = new Thread(r, prefix + "-" + THREAD_COUNTER.incrementAndGet());
+            t.setDaemon(true);
+            return t;
+        };
+    }
+
+    @Nullable
+    private static JsonElement snapshotAttributes(@Nullable EvaluationContext context) {
```
vazarkevych: Quick question on snapshotAttributes() — right now it sends the full user attribute map (plus hash_value) to the ingest endpoint on every event by default. Is that intentional, or would it be safer to make attribute export opt-in (empty by default, with an allowlist), along with a configurable region/host? Given GDPR exposure, wanted to flag it before it lands.

lib/src/main/java/growthbook/sdk/java/plugin/tracking/GrowthBookTrackingPlugin.java
```diff
@@ -0,0 +1,305 @@
...
+
+    public static GrowthBookTrackingPlugin of(TrackingPluginConfig config) {
+        if (config == null) {
+            throw new IllegalArgumentException("TrackingPluginConfig must not be null");
+        }
+        return new GrowthBookTrackingPlugin(config);
+    }
+
+    @Override
+    public void init() {
```
vazarkevych: Noticed that init() doesn't really initialize anything — it just logs the disabled warning, while the real setup (OkHttpClient, scheduler, flush executor) happens in the constructor. That means both executor threads start at construction time, before the plugin is even registered and init() runs, which flips the lifecycle contract on its head — and leaks two daemon executors if the plugin gets constructed but never registered. Would it make sense to move that setup into init() and keep the constructor side-effect-free?

lib/src/main/java/growthbook/sdk/java/plugin/tracking/GrowthBookTrackingPlugin.java
```diff
@@ -0,0 +1,305 @@
...
+            String featureKey,
+            FeatureResult<V> result,
+            EvaluationContext context
+    ) {
+        if (disabled || closed.get()) return;
+        enqueue(TrackingEvent.forFeature(featureKey, result, snapshotAttributes(context)));
+    }
+
+    @Override
+    public void close() {
```
vazarkevych, line 167: close() correctly tears everything down, but it crams roughly five concerns into one body — drain-under-lock, cancel pending, final flush, scheduler shutdown, owned-executor shutdown with awaitTermination, httpClient shutdown — which makes it hard to follow. Extracting each into a small named helper (finalFlush(), shutdownScheduler(), shutdownFlushExecutor(), shutdownHttpClient()) would keep the exact same behavior but turn close() into a clean 4-line sequence where it's obvious nothing gets missed. Worth doing?

lib/src/main/java/growthbook/sdk/java/plugin/tracking/GrowthBookTrackingPlugin.java
```diff
@@ -0,0 +1,305 @@
...
+            } catch (InterruptedException e) {
+                Thread.currentThread().interrupt();
+                es.shutdownNow();
+            }
+        }
+        if (ownsHttpClient) {
+            try {
+                httpClient.dispatcher().executorService().shutdown();
+                httpClient.connectionPool().evictAll();
+            } catch (Throwable t) {
```
vazarkevych: flushBatch() already swallows its own IOException, so the outer catch (Throwable) at each call site (executor task, inline fallback, final flush) really only fires for cases that "can't happen" — which suggests it's over-caught. And Throwable is a bit too broad here, since it also swallows Errors and genuine bugs, logging them as a generic "flush failed" instead of surfacing them. Could we handle the failure once at the source, and if a safety net is still wanted, catch Exception rather than Throwable?

lib/src/main/java/growthbook/sdk/java/plugin/tracking/GrowthBookTrackingPlugin.java
```diff
@@ -0,0 +1,305 @@
...
+                log.debug("Failed to shut down owned OkHttpClient", t);
+            }
+        }
+    }
+
+    // --- internal ---
```
vazarkevych: Small nit: the // --- internal --- divider feels like noise since the private modifiers already make that clear. Could you drop it?

lib/src/main/java/growthbook/sdk/java/plugin/tracking/GrowthBookTrackingPlugin.java
```diff
@@ -0,0 +1,305 @@
...
+                try {
+                    flushBatch(batch);
+                } catch (Throwable t) {
+                    log.warn("Tracking flush failed", t);
+                }
+            });
+        } catch (RejectedExecutionException e) {
+            try {
+                flushBatch(batch);
+            } catch (Throwable t) {
```
vazarkevych: Same pattern here — both catch (Throwable) blocks in submitFlush() (the executor task and the RejectedExecutionException inline fallback) wrap a flushBatch() that already handles its own IOException, so in practice they're only catching bugs and Errors and mislabeling them as "flush failed." Could we narrow those to Exception at most, or better, drop them entirely and let flushBatch own its own outcome?

lib/src/main/java/growthbook/sdk/java/plugin/tracking/GrowthBookTrackingPlugin.java
```diff
@@ -0,0 +1,305 @@
...
+            return t;
+        };
+    }
+
+    @Nullable
+    private static JsonElement snapshotAttributes(@Nullable EvaluationContext context) {
+        if (context == null) return null;
+        UserContext user = context.getUser();
+        if (user == null || user.getAttributes() == null) return null;
+        return user.getAttributes().deepCopy();
```
vazarkevych: One more: snapshotAttributes() runs user.getAttributes().deepCopy() synchronously inside onExperimentViewed/onFeatureEvaluated — so it's on the evaluation thread. Since the plugin contract expects these callbacks to return quickly and not block evaluation, doing a full deep copy of the attributes JSON on every event, on the hot path, seems costly. It also means each buffered event holds a complete copy of the attributes, so a 100-event batch ends up retaining 100 profile copies — and that's amplified further by the unbounded flush queue. Would it be worth capturing only the allowlisted fields instead, and/or moving the copy off the eval thread?

lib/src/main/java/growthbook/sdk/java/plugin/tracking/GrowthBookTrackingPlugin.java
```diff
@@ -0,0 +1,305 @@
...
+
+        if (config.getHttpClient() != null) {
+            this.httpClient = config.getHttpClient();
+            this.ownsHttpClient = false;
+        } else {
+            this.httpClient = new OkHttpClient();
+            this.ownsHttpClient = true;
+        }
+
+        this.scheduler = Executors.newSingleThreadScheduledExecutor(
```
vazarkevych, line 137: One more, related to the earlier init() point: when clientKey is empty the plugin is disabled and every event path early-returns, but the constructor still creates both the scheduler and flush executor. So a disabled plugin ends up spinning up two daemon threads that never do anything. This ties into moving setup into init() — resource creation should probably be skipped entirely when disabled is true.

lib/src/main/java/growthbook/sdk/java/plugin/tracking/TrackingEvent.java
```diff
@@ -0,0 +1,185 @@
...
+                .value(toJson(value))
+                .build();
+    }
+
+    @Nullable
+    private static JsonElement toJson(@Nullable Object value) {
+        if (value == null) return null;
+        try {
+            return GrowthBookJsonUtils.getInstance().gson.toJsonTree(value);
+        } catch (Throwable t) {
```
vazarkevych: Similar issue here — catch (Throwable) is too broad, since it swallows Errors and genuine bugs, not just a serialization failure. Could we narrow it to Exception (or even JsonParseException)? Also, returning null silently drops the value with no trace — it might be worth at least logging at debug level, so a serialization problem is diagnosable instead of showing up as a mysteriously missing value.

lib/src/main/java/growthbook/sdk/java/plugin/tracking/TrackingEvent.java
```diff
@@ -0,0 +1,185 @@
...
+
+    @Nullable
+    @SerializedName("rule_id")
+    private final String ruleId;
+
+    @Nullable
+    @SerializedName("value")
+    private final JsonElement value;
+
+    private TrackingEvent(Builder b) {
```
vazarkevych: In TrackingEvent(Builder b) — any reason for the single-letter b? builder reads more clearly and costs nothing. Same thought applies to the v params on the Builder setters.

lib/src/main/java/growthbook/sdk/java/plugin/tracking/TrackingEvent.java
```diff
@@ -0,0 +1,185 @@
...
+    private static JsonElement toJson(@Nullable Object value) {
+        if (value == null) return null;
+        try {
+            return GrowthBookJsonUtils.getInstance().gson.toJsonTree(value);
+        } catch (Throwable t) {
+            return null;
+        }
+    }
+
+    static final class Builder {
```
vazarkevych: Curious why this uses a hand-written Builder — the rest of the models use Lombok (@Builder/@Data), and this duplicates roughly 40 lines of setter boilerplate without adding any extra behavior. Lombok's @Builder should work fine alongside the @SerializedName fields; using @Builder(access = PRIVATE) (keeping forExperiment/forFeature as the public factories) would drop the manual builder and keep this consistent with the rest of the codebase. What do you think?

lib/src/main/java/growthbook/sdk/java/plugin/tracking/TrackingEvent.java
```diff
@@ -0,0 +1,185 @@
...
+ * A single event dispatched by {@link GrowthBookTrackingPlugin} to the
+ * GrowthBook data-warehouse ingest endpoint. Field names mirror the Go SDK.
+ */
+public final class TrackingEvent {
+
+    public static final String EVENT_EXPERIMENT_VIEWED = "experiment_viewed";
+    public static final String EVENT_FEATURE_EVALUATED = "feature_evaluated";
+
+    @SerializedName("event_type")
+    private final String eventType;
```
vazarkevych: One more small one: eventType is a String with the EVENT_* constants defined alongside it — using an enum instead (with @SerializedName on each value) would be type-safe and catch invalid values at compile time rather than at runtime. Worth the switch?

lib/src/main/java/growthbook/sdk/java/plugin/tracking/TrackingEvent.java
```diff
@@ -0,0 +1,185 @@
...
+import growthbook.sdk.java.util.GrowthBookJsonUtils;
+
+import javax.annotation.Nullable;
+import java.time.Instant;
+
+/**
+ * A single event dispatched by {@link GrowthBookTrackingPlugin} to the
+ * GrowthBook data-warehouse ingest endpoint. Field names mirror the Go SDK.
+ */
+public final class TrackingEvent {
```
vazarkevych: One more: getEventType() is the only getter on this class, while every other field is serialized purely via Gson reflection — why single this one out? If it's just for test convenience, maybe the class should stay a pure serialization DTO and drop the lone getter; otherwise, if accessors are genuinely needed, it might be worth adding them consistently rather than just the one. Which way do you want to go?

lib/src/main/java/growthbook/sdk/java/plugin/tracking/TrackingPluginConfig.java
```diff
@@ -0,0 +1,82 @@
...
+        return batchTimeout == null || batchTimeout.isZero() || batchTimeout.isNegative()
+                ? DEFAULT_BATCH_TIMEOUT : batchTimeout;
+    }
+
+    public Duration resolvedCloseTimeout() {
+        return closeTimeout == null || closeTimeout.isZero() || closeTimeout.isNegative()
+                ? DEFAULT_CLOSE_TIMEOUT : closeTimeout;
+    }
+
+    private static String stripTrailingSlash(String s) {
```
vazarkevych: stripTrailingSlash is a generic string operation living as a private method in a config DTO, but the codebase already has StringUtils (plus endpoint/URL normalization in RemoteEvalRequestBuilder.normalizeUrl) — why not reuse or extend those instead of a local one-off? Same thought for the emptiness check: StringUtils.isBlank(ingestorHost) would also catch whitespace-only hosts, which isEmpty() misses.

lib/src/main/java/growthbook/sdk/java/plugin/tracking/TrackingPluginConfig.java
```diff
@@ -0,0 +1,82 @@
...
+import javax.annotation.Nullable;
+import java.time.Duration;
+import java.util.concurrent.Executor;
+
+/**
+ * Configuration for {@link GrowthBookTrackingPlugin}. Defaults mirror the
+ * GrowthBook Go SDK: POST {@code {ingestorHost}/events}, batch size 100,
+ * flush every 10 seconds.
+ */
+@Getter
```
vazarkevych: `@Getter` here exposes raw nullable getters (`getBatchSize()`, `getIngestorHost()`, …) right alongside the `resolved*()` methods that apply defaults, so each setting ends up with two ways to be read — and the raw one can come back null or 0. That seems easy to misuse (grabbing `getBatchSize()` instead of `resolvedBatchSize()` by mistake). Would it make sense to hide the raw getters where a `resolved*()` exists — `@Getter(AccessLevel.NONE)` on those fields — and expose only the resolved accessors, keeping direct getters just for `clientKey`/`httpClient`/`flushExecutor`?

lib/src/main/java/growthbook/sdk/java/plugin/tracking/TrackingPluginConfig.java
```diff
@@ -0,0 +1,82 @@
...
+     */
+    @Nullable
+    private final Executor flushExecutor;
+
+    public String resolvedIngestorHost() {
+        if (ingestorHost == null || ingestorHost.isEmpty()) return DEFAULT_INGESTOR_HOST;
+        return stripTrailingSlash(ingestorHost);
+    }
+
+    public int resolvedBatchSize() {
```
vazarkevych, line 84: `resolvedBatchSize()` guards against `<= 0` but not an unbounded upper value — a very large `batchSize` means the buffer never eager-flushes and just keeps growing, which compounds the unbounded flush-queue issue mentioned earlier. Might be worth clamping it to a sane max?

lib/src/main/java/growthbook/sdk/java/plugin/PluginRegistry.java
```diff
@@ -0,0 +1,108 @@
...
+/**
+ * Holds a set of {@link GrowthBookPlugin}s registered with a GrowthBook
+ * instance and dispatches lifecycle/event callbacks to each one.
+ *
+ * <p>Dispatch is best-effort: exceptions from one plugin never propagate
+ * to the evaluator or other plugins. {@link #initAll()} captures init failures
+ * so a failing plugin becomes a no-op rather than aborting registration.
+ */
+@Slf4j
+public final class PluginRegistry {
```
vazarkevych, line 23: The `catch (Throwable)` in `initAll`/`fire*`/`closeAll` is a bit too broad. Best-effort isolation across plugins is the right instinct, but `Throwable` also swallows `Error`s (OOM, StackOverflow) and keeps looping regardless — at that point the JVM is likely already dying, and we're just logging a warning instead of letting it surface. Could we catch `Exception` instead, so plugin bugs stay isolated but `Error`s still propagate?

lib/src/main/java/growthbook/sdk/java/plugin/PluginRegistry.java
```diff
@@ -0,0 +1,108 @@
...
+        } else {
+            this.plugins = new ArrayList<>(plugins);
+        }
+    }
+
+    public boolean isEmpty() {
+        return plugins.isEmpty();
+    }
+
+    public void initAll() {
```
vazarkevych, line 45: `initAll()` logs an init failure but still keeps the plugin in the list, so it keeps receiving `onExperimentViewed`/`onFeatureEvaluated`/`close` afterward. That pushes the burden of "staying safe after a failed init" onto every individual plugin (self-guard flags, null checks, etc.). Would it make sense to track init failures and skip those plugins for subsequent dispatch, so a plugin that failed to initialize simply stops receiving events going forward?

lib/src/main/java/growthbook/sdk/java/plugin/PluginRegistry.java
```diff
@@ -0,0 +1,108 @@
...
+ * to the evaluator or other plugins. {@link #initAll()} captures init failures
+ * so a failing plugin becomes a no-op rather than aborting registration.
+ */
+@Slf4j
+public final class PluginRegistry {
+
+    private final List<GrowthBookPlugin> plugins;
+
+    public PluginRegistry(@Nullable List<GrowthBookPlugin> plugins) {
+        if (plugins == null || plugins.isEmpty()) {
```
vazarkevych, line 28: One more small one: null entries in the input list aren't filtered out — a `null` plugin becomes a logged NPE on every dispatch. Cheap fix: filter nulls out in the constructor.

lib/src/main/java/growthbook/sdk/java/plugin/PluginRegistry.java
```diff
@@ -0,0 +1,108 @@
...
+                }
+            } catch (Throwable t) {
+                log.warn("Plugin {} onFeatureEvaluated failed",
+                        plugin.getClass().getName(), t);
+            }
+        }
+    }
+
+    public void closeAll() {
+        for (GrowthBookPlugin plugin : plugins) {
```
vazarkevych, line 94: Small consistency nit: `fire*` methods short-circuit with `if (plugins.isEmpty()) return;`, but `initAll`/`closeAll` don't do the same. Harmless either way, just an inconsistency worth tidying up if you're already in there.

vazarkevych (changes requested):
## Overall

  Nice work — the plugin architecture itself is clean: no-op-default interface, a
  best-effort `PluginRegistry` that isolates plugin failures, sensible no-op degradation,
  and genuinely good test coverage (real `MockWebServer`, batch/timer/close/failure paths,
  both single- and multi-user wiring). Most of my line comments are quality/consistency
  nits. There are, however, three things I'd treat as blockers and a recurring theme worth
  addressing in one pass.

  ### Must-fix
  
  1. **Data privacy (default attribute export).** `snapshotAttributes()` ships the full
     user attribute map — plus `hash_value` and the feature/experiment `value` — to the
     ingest endpoint (US default host) on every event, opt-out rather than opt-in. For
     GDPR users this makes the plugin hard to enable safely. Please make attribute export
     opt-in (allowlist, empty by default) with a configurable region/host, and document
     that enabling the plugin transfers user data to a third party.

  2. **Remote-eval path is behind `main` and re-fires exposures.** The inline
     `trackData.forEach(...)` loop predates #216, which replaced it with
     `experimentEvaluator.fireRemoteEvaluationTracks(...)` (de-dupes per assignment,
     null-guards payloads). As written, plugins get duplicate `experiment_viewed` on cached
     remote-eval responses, unlike the local `isExperimentTracked` path. Rebase and route
     plugin exposure through that shared, de-duplicated path.

  3. **Memory/latency under load.** The owned flush executor uses an unbounded queue, and
     `feature_evaluated` fires on every evaluation, so a slow/failing endpoint lets batches
     pile up unbounded. On top of that, `snapshotAttributes()` does a full `deepCopy()` on
     the evaluation thread and every buffered event retains its own copy. Needs a bounded
     queue + drop policy, and the attribute capture off the hot path (an allowlist also
     shrinks this a lot).

  ### Recurring theme — keep the evaluators small, catch narrowly

  - The plugin hooks are inlined into the already-huge `evaluateExperiment` /
    `evaluateFeature`, even though this PR already introduced `dispatchFeatureUsage(...)`
    in the same file. Add a `dispatchExperimentViewed(...)` helper and route all three
    sites through helpers so the methods shrink rather than grow — and trim the redundant
    "what" comments while you're touching those blocks.
  - `catch (Throwable)` is used broadly (plugin flush, `submitFlush`, `TrackingEvent.toJson`,
    `PluginRegistry`). Several are redundant (`flushBatch` already swallows its
    `IOException`), and `Throwable` hides `Error`s/bugs. Narrow to `Exception`, handle each
    failure once, and log rather than silently dropping.

  ### Lifecycle / API

  - `init()` does nothing; all setup (threads, http client) is in the constructor, so a
    disabled or never-registered plugin still spins up executors. Move setup into `init()`
    and keep the constructor side-effect-free.
  - `close()` bundles ~5 concerns — extract `finalFlush()/shutdownScheduler()/…`. On the
    `GrowthBook` side, the new `close()` should probably implement `AutoCloseable`, and the
    `if (pluginRegistry != null)` guards on a `final` field are dead.
  - `TrackingPluginConfig` exposes both raw nullable getters and `resolved*()` — easy to
    grab the wrong one; hide the raw ones. `batchSize` has no upper clamp.

  ### Nits

  `ValueType` vs idiomatic `<V>`; hand-rolled Builder vs the Lombok `@Builder` used
  elsewhere; `eventType` String vs enum; `stripTrailingSlash`/`isEmpty` duplicating
  `StringUtils`/`RemoteEvalRequestBuilder.normalizeUrl`; `// --- internal ---` divider;
  single-letter param names; per-event `sdk_language`/`sdk_version` that could live on the
  batch envelope.
  
  
We completely understand if you're busy at the moment. If it's helpful, we'd be happy to work on these improvements ourselves and open PRs for you to review when you have time.

lib/src/main/java/growthbook/sdk/java/evaluators/ExperimentEvaluator.java
madhuchavva: Done in 13a92e5 — extracted `dispatchExperimentViewed(context, experiment, result)`, used by both the local and remote-eval paths, so the method gains a single call instead of a new null-check block, and I trimmed the restated comments in that block.

lib/src/main/java/growthbook/sdk/java/evaluators/FeatureEvaluator.java
madhuchavva: Kept `ValueType` here deliberately: the whole evaluator layer (ExperimentEvaluator/FeatureEvaluator signatures) uses `<ValueType>`, so switching just this helper to `V` would make the file inconsistent. Happy to do a repo-wide rename to the single-letter convention as a separate cleanup PR if you'd prefer.

lib/src/main/java/growthbook/sdk/java/evaluators/FeatureEvaluator.java
madhuchavva: Trimmed the pure-restatement comments around the dispatch sites in 13a92e5. Left the ones carrying non-obvious intent (rollout / prerequisite handling) so the diff stays scoped to this PR rather than churning unrelated lines.

lib/src/main/java/growthbook/sdk/java/evaluators/FeatureEvaluator.java
madhuchavva: Resolved — the inline block is gone. Remote-eval tracking now routes through `experimentEvaluator.fireRemoteEvaluationTracks(...)`, and that helper dispatches to plugins as well (see the next comment).

lib/src/main/java/growthbook/sdk/java/evaluators/FeatureEvaluator.java
madhuchavva: Done. Rebased onto #216: the parallel `trackData.forEach` loop is deleted and plugin `fireExperimentViewed` now goes through `fireRemoteEvaluationTracks`, deduped via the shared `alreadyTracked` guard. Note the branch's old code used `t.getResult().getExperimentResult()`, which is wrong against main's `TrackData` (its `result` is already the `ExperimentResult`) — the merged code uses `track.getResult()`.

lib/src/main/java/growthbook/sdk/java/plugin/tracking/GrowthBookTrackingPlugin.java
madhuchavva: Resolved by dropping attribute export entirely: the Go SDK sends no user attributes, so to match its wire contract `snapshotAttributes()` is removed and events no longer carry an `attributes` map (or the raw profile). The ingest host is already configurable via `ingestorHost`.

lib/src/main/java/growthbook/sdk/java/plugin/tracking/GrowthBookTrackingPlugin.java
madhuchavva: Fixed — the HTTP client, scheduler, and flush executor are now created in `init()`; the constructor only stores config and computes `disabled`, so it is side-effect-free.

lib/src/main/java/growthbook/sdk/java/plugin/tracking/GrowthBookTrackingPlugin.java
madhuchavva: Done — `close()` is now a 4-line sequence delegating to `finalFlush()`, `shutdownScheduler()`, `shutdownFlushExecutor()`, and `shutdownHttpClient()`.

lib/src/main/java/growthbook/sdk/java/plugin/tracking/GrowthBookTrackingPlugin.java
madhuchavva: Handled at the source now: `flushBatch` owns its outcome with a single `catch (Exception)` (so `Error`s propagate), and the outer `catch (Throwable)` blocks at the call sites are removed.

lib/src/main/java/growthbook/sdk/java/plugin/tracking/GrowthBookTrackingPlugin.java
madhuchavva: Removed.

lib/src/main/java/growthbook/sdk/java/plugin/tracking/GrowthBookTrackingPlugin.java
madhuchavva: Removed — `submitFlush` no longer wraps `flushBatch` in a try/catch. The owned executor uses a bounded queue and drops the newest batch (with a warning) when full; a user-supplied executor that rejects is caught and the batch dropped the same way.

lib/src/main/java/growthbook/sdk/java/plugin/tracking/GrowthBookTrackingPlugin.java
madhuchavva: Removed entirely — no attribute copy on the eval thread, and events no longer retain attribute copies. Combined with dropping attribute export and the new bounded flush queue (drop-newest, cap 256), the per-batch retention concern is gone.

lib/src/main/java/growthbook/sdk/java/plugin/tracking/GrowthBookTrackingPlugin.java
madhuchavva: Fixed — when `clientKey` is blank, `init()` logs and early-returns before creating any executors, so a disabled plugin spins up no threads.

lib/src/main/java/growthbook/sdk/java/plugin/tracking/TrackingEvent.java
madhuchavva: Narrowed to `catch (Exception)` and it now logs at debug before dropping the value, so a serialization problem is diagnosable.

lib/src/main/java/growthbook/sdk/java/plugin/tracking/TrackingEvent.java
madhuchavva: Moot now — `TrackingEvent` uses Lombok `@Builder`, so the hand-written builder and its single-letter params are gone.

lib/src/main/java/growthbook/sdk/java/plugin/tracking/TrackingEvent.java
madhuchavva: Done — switched to Lombok `@Builder(access = AccessLevel.PRIVATE)` with `forExperiment`/`forFeature` as the public factories, dropping ~40 lines of manual builder boilerplate.

lib/src/main/java/growthbook/sdk/java/plugin/tracking/TrackingEvent.java
madhuchavva: Done — `event_type` is now an `EventType` enum with `@SerializedName` on each constant.

lib/src/main/java/growthbook/sdk/java/plugin/tracking/TrackingEvent.java
madhuchavva: Dropped — `TrackingEvent` is now a pure serialization DTO with no getters (Gson serializes the fields directly).

lib/src/main/java/growthbook/sdk/java/plugin/tracking/TrackingPluginConfig.java
madhuchavva: Now uses `StringUtils.isBlank` for the emptiness check (also catches whitespace-only hosts). Kept a local trailing-slash strip since `RemoteEvalRequestBuilder.normalizeUrl` only null-guards and does not strip slashes — happy to promote a shared helper into `StringUtils` if you'd prefer.

lib/src/main/java/growthbook/sdk/java/plugin/tracking/TrackingPluginConfig.java
madhuchavva: Done — `@Getter(AccessLevel.NONE)` on the fields that have `resolved*()` accessors, so only the resolved values are readable. Direct getters are kept just for `clientKey` / `httpClient` / `flushExecutor`.

lib/src/main/java/growthbook/sdk/java/plugin/tracking/TrackingPluginConfig.java
madhuchavva: Done — `resolvedBatchSize()` now clamps to `MAX_BATCH_SIZE` (10,000), so an oversized value can no longer defeat eager flushing. Paired with the new bounded flush queue.

lib/src/main/java/growthbook/sdk/java/plugin/PluginRegistry.java
madhuchavva: Narrowed to `catch (Exception)` across `initAll`/`fire*`/`closeAll`, so plugin bugs stay isolated but `Error`s propagate.

lib/src/main/java/growthbook/sdk/java/plugin/PluginRegistry.java
madhuchavva: Done — `initAll()` now drops any plugin whose `init()` throws, so it receives no further events and no longer has to self-guard. Added a regression test (`pluginWhoseInitFailsReceivesNoFurtherEvents`).

lib/src/main/java/growthbook/sdk/java/plugin/PluginRegistry.java
madhuchavva: Done — null entries are filtered out in the constructor. Added a test (`nullPluginEntriesAreFilteredOut`).

lib/src/main/java/growthbook/sdk/java/plugin/PluginRegistry.java
madhuchavva: Done — `initAll`/`closeAll` now short-circuit on an empty plugin list, matching `fire*`.

PR #214, opened 2026-06-16
fix: prevent unhandled exceptions when processing empty SSE event payloads

vazarkevych:
### Overview

This PR prevents unhandled exceptions when SSE events contain an empty or heartbeat payload.

Some SSE servers may send keepalive, heartbeat, ping, or empty `data` events. Previously, those payloads could still reach feature JSON parsing and fail when the payload was blank or invalid.

### What Changed

- Added SSE payload validation before feature payload processing.
- Ignored empty and whitespace-only SSE payloads.
- Ignored heartbeat-style SSE events: `heartbeat`, `keepalive`, and `ping`.
- Caught parsing/runtime errors inside the async SSE listener so they do not escape as unhandled exceptions.
- Updated native Java SSE handling to skip whitespace-only payloads.
- Added tests for empty payloads, valid payloads, invalid payload handling, and heartbeat events.

### DIFF
--- a/lib/src/main/java/growthbook/sdk/java/repository/NativeJavaGbFeatureRepository.java
+++ b/lib/src/main/java/growthbook/sdk/java/repository/NativeJavaGbFeatureRepository.java
@@ -4,7 +4,6 @@ import com.google.gson.JsonElement;
 import com.google.gson.JsonObject;
 import growthbook.sdk.java.model.Feature;
 import growthbook.sdk.java.multiusermode.util.TransformationUtil;
-import growthbook.sdk.java.sandbox.FileCachingManagerImpl;
 import growthbook.sdk.java.sandbox.GbCacheManager;
 import growthbook.sdk.java.util.DecryptionUtils;
 import growthbook.sdk.java.exception.FeatureFetchException;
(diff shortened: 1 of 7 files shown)


lib/src/main/java/growthbook/sdk/java/repository/NativeJavaGbFeatureRepository.java
```diff
@@ -531,7 +530,7 @@ public void run() {
                             dataBuffer.append(line.substring(QUANTITY_TO_CUT_SSE).trim()).append("\n");
                         } else if (line.isEmpty()) {
                             String data = dataBuffer.toString();
-                            if (!data.isEmpty()) {
+                            if (!data.trim().isEmpty()) {
```
madhuchavva: [P2] Apply the SSE payload validation to the native reader too

The new `SseEventPayloadValidator` is only used by the OkHttp SSE listener. This native SSE reader still ignores `event:` lines and sends every non-blank `data:` block to `onResponseJson`, so a non-empty heartbeat/keepalive payload can still be parsed as feature JSON in this path. Please share the same event/payload validation here, or narrow the fix to only the OkHttp implementation.

madhuchavva (commented):
Native SSE validation path issue found.

madhuchavva (changes requested):
Requesting changes for the native SSE validation gap noted inline.

madhuchavva (approved):

PR #215, opened 2026-06-16
feat: add feature refresh listeners

vazarkevych:
### Summary

This update adds feature refresh listener support to the Java SDK.

Consumers can now subscribe to feature refresh events and react when feature definitions are refreshed successfully or when a refresh attempt fails. The event payload is metadata-only and does not expose feature definitions directly.

This allows SDK users to observe refresh lifecycle events, trigger custom logging or monitoring, and integrate refresh status with their own application workflows.

---

#### Feature Refresh Listener API

Added client-level APIs for managing feature refresh listeners:

1. `addFeatureRefreshListener(...)`
2. `removeFeatureRefreshListener(...)`
3. `subscribeFeatureRefreshListener(...)`

The `subscribeFeatureRefreshListener(...)` API returns a `FeatureRefreshSubscription`, which provides an idempotent cleanup mechanism for unsubscribing listeners.

---

#### Feature Refresh Event Metadata

Added `FeatureRefreshEvent` to provide metadata about feature refresh results.

Listeners can now receive events when:

1. Feature refresh completes successfully.
2. Feature refresh fails.
3. Refresh metadata needs to be inspected by the consumer.

The event is intentionally metadata-only, so listener consumers can react to refresh status without directly depending on feature payload internals.

---

#### Listener Dispatch Support

Feature refresh listener dispatch is now supported across the refresh flow, including:

1. `GrowthBookClient`
2. `GBFeaturesRepository`
3. `NativeJavaGbFeatureRepository`
4. `LocalGbFeatureRepository`

This ensures refresh events are propagated consistently across client-level and repository-level refresh operations.

---

#### Listener Failure Isolation

Listener execution is isolated from the SDK refresh flow.

If a listener throws an exception:

1. The exception does not break the feature refresh process.
2. Other listeners can still be executed.
3. SDK feature fetching and cache behavior remain unaffected.

This prevents consumer-side listener failures from impacting core SDK behavior.

---

#### Optional Listener Executor

Added optional executor support through:

```java
Options.featureRefreshListenerExecutor(...)
````

This allows consumers to control how listener callbacks are dispatched, including custom threading or asynchronous execution strategies.

---

#### Repository Initialization Refactor

Refactored `GrowthBookClient` repository initialization into an internal provider.

This keeps repository creation and lifecycle handling more centralized, while preserving the public SDK behavior.

---

### DIFF
--- a/lib/src/main/java/growthbook/sdk/java/evaluators/ConditionEvaluator.java
+++ b/lib/src/main/java/growthbook/sdk/java/evaluators/ConditionEvaluator.java
@@ -4,7 +4,6 @@ import com.google.gson.JsonArray;
 import com.google.gson.JsonElement;
 import com.google.gson.JsonObject;
 import com.google.gson.JsonPrimitive;
-import com.google.gson.JsonSyntaxException;
 import growthbook.sdk.java.util.GrowthBookJsonUtils;
 import growthbook.sdk.java.model.Operator;
 import growthbook.sdk.java.util.StringUtils;
@@ -12,15 +11,10 @@ import growthbook.sdk.java.model.DataType;
 import lombok.extern.slf4j.Slf4j;
 
 import javax.annotation.Nullable;
-import java.util.ArrayList;
-import java.util.Collections;
-import java.util.Map;
 import java.util.Objects;
-import java.util.Set;
 import java.util.function.Function;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
-import java.util.regex.PatternSyntaxException;
 
 /**
  * <b>INTERNAL</b>: Implementation of condition evaluation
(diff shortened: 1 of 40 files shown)

### CODE REVIEW

File: lib/src/main/java/growthbook/sdk/java/multiusermode/internal/GlobalContextManager.java [hunk 1]
```diff
@@ -0,0 +1,98 @@
...
+        this.globalContext.set(createGlobalContext(repository));
+    }
+
+    /**
+     * Creates an evaluation context for a specific user by overlaying user attributes on global attributes.
+     *
+     * @param userContext per-user evaluation context
+     * @return evaluation context used by feature and experiment evaluators
+     */
+    public EvaluationContext createEvaluationContext(UserContext userContext) {
```
madhuchavva, line 53: [P0] This drops the sticky-bucket assignment preload that GrowthBookClient currently performs before local evaluations. The old toUserContextWithMergedAttributes path fetched stickyBucketService assignments when the user context had no stickyBucketAssignmentDocs; this refactor only merges attributes and builds the EvaluationContext. That changes experiment assignment behavior for multi-user clients using StickyBucketService and is inconsistent with the existing Java contract and the sticky-bucketing behavior expected across SDKs. Please preserve the sticky assignment loading path and add a regression test that evaluates through GrowthBookClient with a configured StickyBucketService.

lib/src/main/java/growthbook/sdk/java/evaluators/ConditionEvaluator.java
```diff
@@ -351,80 +177,106 @@ Boolean evalOperatorCondition(String operatorString, @Nullable JsonElement actua
...
+
+    /**
+     * {@code $gt}/{@code $gte}/{@code $lt}/{@code $lte}: numeric or lexical comparison.
+     */
+    private Boolean evalComparison(Operator operator, @Nullable JsonElement actual, JsonElement expected, DataType attributeDataType) {
+        if (actual == null || DataType.NULL.equals(attributeDataType)) {
+            if (expected.isJsonPrimitive() && !expected.getAsJsonPrimitive().isNumber()) {
+                return false;
+            }
+            return matchesSign(operator, Double.compare(0.0, expected.getAsDouble()));
```
madhuchavva, line 208: [P1] This changes missing/null numeric comparisons from false to comparing the missing value as 0. For example, a condition like {"age":{"$lt":5}} now matches a user with no age attribute because Double.compare(0.0, 5) is negative. That is an evaluation behavior change and should stay aligned with cases.json / JS / Python semantics. Please keep missing attributes false for $gt/$gte/$lt/$lte unless there is an explicit cross-SDK test case proving this new behavior is intended.

lib/src/main/java/growthbook/sdk/java/multiusermode/internal/RemoteEvalCoordinator.java
```diff
@@ -0,0 +1,326 @@
...
+
+    private UserContext withMergedAttributes(UserContext userContext) {
+        JsonObject merged = new JsonObject();
+        if (this.options.getGlobalAttributes() != null) {
+            merged = GrowthBookJsonUtils.getInstance().gson.fromJson(this.options.getGlobalAttributes(), JsonObject.class);
+            if (merged == null) {
+                merged = new JsonObject();
+            }
+        }
+        JsonObject userAttributes = userContext.getAttributes();
```
madhuchavva: [P1] preloadRemoteEval(null) regresses here. GrowthBookClient.preloadRemoteEval passes the caller-provided context straight into RemoteEvalCoordinator.preload, and this helper immediately dereferences userContext. The previous GrowthBookClient merge helper treated null as UserContext.builder().build(), so this is a compatibility break for callers using null to warm the default/global remote-eval response. Please restore the null-safe default and add a regression test for preloadRemoteEval(null).

lib/src/main/java/growthbook/sdk/java/multiusermode/configurations/Options.java
```diff
@@ -102,19 +99,21 @@ public Options(@Nullable Boolean enabled,
...
                    @Nullable Integer swrTtlSeconds,
-                   @Nullable FeatureRefreshCallback featureRefreshCallback,
+                   @Deprecated @Nullable FeatureRefreshCallback featureRefreshCallback,
                    @Nullable JsonObject globalAttributes,
                    @Nullable Map<String, Object> globalForcedFeatureValues,
                    @Nullable Map<String, ?> globalForcedVariationsMap,
                    @Nullable GbCacheManager cacheManager,
                    @Nullable CacheMode cacheMode,
                    @Nullable String cacheDirectory,
+                   @Nullable Executor featureRefreshListenerExecutor,
```
madhuchavva: [P1] Adding featureRefreshListenerExecutor in the middle of the public @Builder constructor changes the direct-constructor signature/order for the options added in the remote-eval/refresh work. Existing callers using the long public constructor with remoteEval/cacheKeyAttributes/remoteEvalCacheSize/remoteEvalCacheTtlSeconds/backgroundFetchInterval/retryPolicy after cacheDirectory will no longer compile. Since this PR is about refresh listeners, please preserve the existing constructor signature and add any new parameter at the end or use an overload so builder users get the new option without breaking direct constructor users.

madhuchavva (changes requested):
Requesting changes. This PR goes beyond feature-refresh listeners and currently includes functional regressions: sticky-bucket assignment preload is dropped from GrowthBookClient local evaluations, missing numeric comparisons change evaluation behavior, preloadRemoteEval(null) now dereferences null, and the public Options constructor parameter order changes. Please narrow/preserve behavior and add focused regression tests around these paths.

PR #216, opened 2026-06-16
feat: add remote evaluation support

vazarkevych:
## Summary

Adds remote evaluation support to the Java SDK for both single-context `GrowthBook` usage and multi-user `GrowthBookClient`.

## Changes

- Added remote eval configuration options: `remoteEval`, `apiHost`, `clientKey`, `cacheKeyAttributes`, `remoteEvalCacheSize`, and `remoteEvalCacheTtlSeconds`.
- Added remote eval HTTP service, response parser, endpoint helper, request builder, deterministic cache key builder, and in-memory cache.
- Integrated remote eval into feature evaluation with safe fallback to local/current feature context when remote fetch fails.
- Added remote eval response caching with TTL, stale-while-revalidate support, cache invalidation, and shutdown cleanup.
- Added SSE-based remote eval cache invalidation for multi-user mode.
- Added option validation for unsupported remote eval combinations, including encrypted payloads, sticky bucketing, SWR refresh strategy, and GrowthBook Cloud host usage.
- Updated remote tracking handling so remotely evaluated experiment tracks are deduplicated per assignment.
- Normalized remote eval request payload defaults to avoid null fields.
- Fixed condition evaluator behavior for `$ne`, `$notRegex`, and `$notRegexi` with missing/null attributes.

### DIFF
--- a/lib/src/main/java/growthbook/sdk/java/multiusermode/GrowthBookClient.java
+++ b/lib/src/main/java/growthbook/sdk/java/multiusermode/GrowthBookClient.java
@@ -41,6 +41,7 @@ import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Objects;
+import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.concurrent.atomic.AtomicReference;
 
 @Slf4j
@@ -55,8 +56,9 @@ public class GrowthBookClient {
     private final ExperimentEvaluator experimentEvaluatorEvaluator;
     private final AtomicReference<GlobalContext> globalContext = new AtomicReference<>();
     private final AtomicReference<GBFeaturesRepository> repository = new AtomicReference<>();
-    private RemoteEvalService remoteEvalService;
-    private RemoteEvalCache remoteEvalCache;
+    private volatile RemoteEvalService remoteEvalService;
+    private volatile RemoteEvalCache remoteEvalCache;
+    private final AtomicBoolean remoteEvalReady = new AtomicBoolean(false);
 
     public GrowthBookClient() {
         this(Options.builder().build());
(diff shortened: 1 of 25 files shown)


lib/src/main/java/growthbook/sdk/java/multiusermode/GrowthBookClient.java
```diff
@@ -246,29 +338,140 @@ public void onError(Throwable throwable) {
...
-                    .map(entry -> Arrays.asList(entry.getKey(), entry.getValue()))
-                    .collect(Collectors.toList());
+    private RemoteEvalResponse getRemoteEvalResponse(UserContext userContext) throws FeatureFetchException {
+        ensureRemoteEvalReady();
+
+        Map<String, Integer> forcedVariations = mergeForcedVariations(userContext);
+        Map<String, Object> forcedFeatures = mergeForcedFeatures(userContext);
+        String url = userContext.getUrl() == null ? this.options.getUrl() : userContext.getUrl();
+        url = RemoteEvalRequestBuilder.normalizeUrl(url);
+        String cacheKey = RemoteEvalCacheKey.fromContext(
```
madhuchavva, line 489: [P1] Include forced feature overrides in the remote-eval cache key

`forcedFeatures` is merged and sent in the request body below, but it is not part of the cache key. That means two evaluations for the same user/url/forced variations but different `UserContext.forcedFeatureValues` can reuse the first cached remote-eval response and return a result computed with the wrong overrides. The same omission exists in the single-user `GrowthBook` path. Please include a canonicalized forced-features payload in `RemoteEvalCacheKey`, or skip cache reuse when forced feature values differ.

madhuchavva (commented):
Remote-eval cache key issue found.

madhuchavva (changes requested):
Requesting changes for the remote-eval cache-key issue noted inline.

madhuchavva (approved):

PR #218, opened 2026-06-26
Fix!: Update OkHttp dependencies to version 5.4.0

mringholm:
This pull request updates the versions of the OkHttp and OkHttp-SSE dependencies to ensure the project uses the latest features and security patches.

Dependency updates:

* Upgraded `okhttp` from version 4.11.0 to 5.4.0 in `lib/build.gradle`.
* Upgraded `okhttp-sse` from version 4.11.0 to 5.4.0 in `lib/build.gradle`.

### DIFF
--- a/lib/build.gradle
+++ b/lib/build.gradle
@@ -53,9 +53,9 @@ dependencies {
 
 
     // https://square.github.io/okhttp
-    implementation 'com.squareup.okhttp3:okhttp:4.11.0'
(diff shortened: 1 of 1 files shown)


madhuchavva (approved):
Approved. OkHttp 5 is dependency-sensitive and should remain marked as breaking/release-noted, but the SDK compiles and the full :lib:test suite passed locally against this branch.

PR #221, opened 2026-07-03
feat: add JCache (JSR-107) cache adapter module

vazarkevych:
# Add JCache (JSR-107) cache adapter (`growthbook-cache-jcache`)

## Summary

Adds a new optional Gradle module, `growthbook-cache-jcache`, that provides a JSR-107 (JCache) backed implementation of the SDK's `GbCacheManager`. This lets applications plug the GrowthBook feature cache into any standard JCache provider (Ehcache, Hazelcast, Caffeine-JCache, the JSR-107 reference implementation, etc.) instead of keeping it in-process or on the local filesystem.

The adapter depends only on the `javax.cache:cache-api` interfaces — the consumer brings their own provider on the runtime classpath.

## Design

- **Provider-owned semantics.** Unlike an embedded cache, a JSR-107 cache is created and configured by its provider, so the adapter is given a `CacheManager` plus the name of the cache to use. Eviction, expiry, and sizing are the provider's responsibility and are configured on the provider side — the adapter only stores feature payloads with their last-updated timestamp.
- **Storage format.** Each entry is a `JCacheCacheEntry` holding the payload and an `updatedAt` (epoch millis) timestamp. It implements `Serializable` because JSR-107 caches default to store-by-value semantics, where the provider copies values via serialization. `getLastUpdatedMillis()` is backed by that timestamp, enabling the SDK's freshness-based refresh skipping.
- **Cache resolution.** `resolveCache` reuses an existing cache when present, otherwise creates one with a typed `MutableConfiguration`. A concurrent-creation race (another caller created the cache first) is handled by re-fetching the cache rather than failing.
- **Configuration** via a builder (`JCacheCacheOptions`):
  - `cacheManager` — the JSR-107 cache manager used to obtain (or create) the feature cache. Required.
  - `cacheName` — name of the cache within the manager (default `growthbook-features`).
  - `clock` — injectable for testing.
- **Error handling.** An absent key is a normal cache miss (`null`); genuine access failures are wrapped in `FeatureCacheException` with the offending key for context.

## Usage

```java
GbCacheManager cache = JCacheGbCacheManager.builder()
        .cacheManager(cacheManager)      // your JSR-107 CacheManager
        .cacheName("growthbook-features")
        .buildManager();

// Or, with defaults:
GbCacheManager cache = JCacheGbCacheManager.create(cacheManager);
```

## Testing

- **Unit tests** (`JCacheGbCacheManagerTest`) run against the Ehcache JSR-107 provider (`testRuntimeOnly`), exercising save/load, last-updated tracking, cache miss, and clear behaviour.

## Build / packaging

- New module registered in `settings.gradle`.
- Targets Java 8 (`sourceCompatibility`/`targetCompatibility = 1.8`).
- Depends on `javax.cache:cache-api:1.1.1` (as `api`); a provider is required only at runtime.
- Publishes `sources` and `javadoc` jars to GitHub Packages, consistent with the existing `lib` module.

### DIFF
--- a/growthbook-cache-jcache/build.gradle
+++ b/growthbook-cache-jcache/build.gradle
@@ -1,0 +1,73 @@
+plugins {
+    id 'java-library'
+    id 'maven-publish'
+}
+
+sourceCompatibility = JavaVersion.VERSION_1_8
+targetCompatibility = JavaVersion.VERSION_1_8
+
+repositories {
+    mavenCentral()
+}
+
+dependencies {
+    api project(':lib')
+
+    api 'javax.cache:cache-api:1.1.1'
+
+    testImplementation 'org.junit.jupiter:junit-jupiter:5.8.2'
+    // A JSR-107 provider implementation is required to run the tests.
+    testRuntimeOnly 'org.ehcache:ehcache:3.10.8'
+}
+
+java {
+    withSourcesJar()
+    withJavadocJar()
+}
+
+tasks.named('test') {
+    useJUnitPlatform()
+}
+
+publishing {
+    publications {
+        mavenJava(MavenPublication) {
+            from components.java
+
+            pom {
+                name = 'GrowthBook SDK Java JCache Cache Adapter'
+                description = 'Optional JSR-107 (JCache) cache adapter for the GrowthBook Java SDK'
+                url = 'https://github.com/growthbook/growthbook-sdk-java'
+                licenses {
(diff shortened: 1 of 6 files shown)


madhuchavva (approved):
Approved. The JCache adapter is additive, keeps provider-owned expiry/sizing semantics, and the module tests passed locally.

PR #222, opened 2026-07-03
feat: Add Redis cache adapter

vazarkevych:
---

# Add Redis cache adapter (growthbook-cache-redis)

## Summary

Adds a new optional Gradle module, `growthbook-cache-redis`, that provides a Redis-backed implementation of the SDK's `GbCacheManager`. This enables sharing the GrowthBook feature cache across multiple instances/pods instead of keeping it in-process or on the local filesystem.

Two Redis clients are supported out of the box:
- **Jedis** — `JedisGbCacheManager`
- **Lettuce** — `LettuceGbCacheManager`

Both client dependencies are `compileOnly`, so consumers include only the client they actually use; nothing is forced onto the SDK's runtime classpath.

---

### Shared base, thin adapters
`AbstractRedisGbCacheManager` contains the shared logic:
- key namespacing
- Redis hash ↔ entry mapping
- cache-miss vs failure handling
- `FeatureCacheException` wrapping

Each adapter only implements low-level Redis operations:
- `writeHash`
- `readHash`
- `deleteByPrefix`

---

### Storage format
Each cache entry is stored as a Redis hash with:
- `data`
- `updatedAt` (epoch millis)

`RedisCacheEntry` centralizes field names and mapping so both adapters use the same on-the-wire format.

`getLastUpdatedMillis()` is backed by `updatedAt`, enabling freshness-based cache refresh skipping in the SDK.

---

### Configuration (self-typed builder)
`AbstractRedisCacheOptions` provides configuration:

- **keyPrefix**  
  Namespace for all keys (default: `growthbook:features:`).  
  `clearCache()` removes only keys under this prefix via `SCAN` + `DEL` (no blocking `KEYS`).

- **ttl**  
  Optional per-entry TTL applied on write. `null` means no expiration.

- **clock**  
  Injectable clock for deterministic testing.

---

### Connection ownership
The caller owns the lifecycle of:
- `JedisPool`
- `StatefulRedisConnection`

The adapter only borrows connections and executes commands.

---

### Error handling
- Missing key → treated as cache miss (`null`)
- Real Redis failures → wrapped in `FeatureCacheException` with the affected key for context

---

### Jedis

```java
GbCacheManager cache = JedisGbCacheManager.builder()
        .jedisPool(jedisPool)
        .keyPrefix("growthbook:features:")
        .ttl(Duration.ofMinutes(10))
        .buildManager();
```
## Lettuce

```java
GbCacheManager cache = LettuceGbCacheManager.builder()
        .connection(connection)
        .buildManager();
```

---

### Unit tests
- JedisGbCacheManagerTest
- LettuceGbCacheManagerTest
- Implemented with Mockito

### Integration tests
- RedisCacheIntegrationTest
- Runs against real Redis via Testcontainers
- Located in a separate `integrationTest` source set
- Exposed via opt-in `integrationTest` Gradle task
- Not included in `check`, so standard builds do not require Docker

---

## Build / packaging

- New module registered in `settings.gradle`
- Targets Java 8 (`sourceCompatibility` / `targetCompatibility = 1.8`)
- Publishes sources and Javadoc jars to GitHub Packages
- Fully consistent with existing `lib` module


growthbook-cache-redis/src/main/java/growthbook/sdk/java/cache/redis/AbstractRedisGbCacheManager.java
```diff
@@ -0,0 +1,125 @@
...
+    protected static final int SCAN_BATCH_SIZE = 256;
+
+    private final Clock clock;
+    private final Long ttlSeconds;
+    private final String keyPrefix;
+
+    protected AbstractRedisGbCacheManager(String keyPrefix, Duration ttl, Clock clock) {
+        this.keyPrefix = keyPrefix;
+        this.clock = clock;
+        this.ttlSeconds = ttl == null ? null : ttl.getSeconds();
```
madhuchavva: [P1] The options builder accepts any positive Duration, but this conversion truncates sub-second TTLs to 0. Passing Duration.ofMillis(500) therefore writes EXPIRE key 0 in both Jedis and Lettuce paths, which immediately expires/deletes the cache entry even though the caller configured a positive TTL. Please either use millisecond precision (PEXPIRE / pexpire) or reject/round sub-second TTLs explicitly, and add a regression test for a sub-second TTL.

madhuchavva (changes requested):
Requesting changes for the Redis TTL precision issue. The module tests pass, but accepting positive sub-second Duration values and truncating with getSeconds() can immediately expire/delete Redis entries. Please fix the TTL handling and add a regression test.

PR #223, opened 2026-07-03
feat: add configuration validation at client start-up

vazarkevych:
## Summary
  
  Adds start-up **configuration validation** for the multi-user `GrowthBookClient`.
  Invalid or contradictory options now fail at `initialize()` with a single, clear,
  aggregated message instead of surfacing later as an opaque feature-fetch failure.

  ## What changed

  - **`OptionsValidator`** (new) — validates `Options` and is invoked at the very
    start of `GrowthBookClient.initialize()`. Checks:
    - `apiHost` — present and a syntactically valid `http(s)` URL (no network call)
    - `clientKey` — present
    - `swrTtlSeconds` (refresh interval) — positive when set
    - `backgroundFetchInterval` — non-negative when set
    - `remoteEvalCacheTtlSeconds` — positive when set
    - cache config conflicts — `cacheManager` supplied while caching is disabled;
      `CacheMode.CUSTOM` without a `cacheManager`
    - remote-eval incompatibilities (delegated, see below)
  - **`InvalidOptionsException`** (new) — extends `IllegalArgumentException` and
    carries the full `List<String> getViolations()` for structured consumers
    (upcoming diagnostics / health checks).
  - **`RemoteEvalOptionsValidator`** (refactor) — converted to a collecting model;
    `validate(...)` now aggregates instead of failing on the first problem, and a
    new `remoteEvalViolations(Options)` lets `OptionsValidator` merge remote-eval
    constraints through **one** validation entry point (no duplicate
    `apiHost`/`clientKey` messages).
  - **Errors are aggregated** — all problems are reported together, so a caller
    fixing several misconfigured options sees every issue at once.

  ## Behaviour & compatibility
  
  - `GrowthBookClient.initialize()` keeps its `boolean` contract: validation
    failures are logged and return `false` (the exception is caught internally).
  - `InvalidOptionsException` **is** an `IllegalArgumentException`, so existing
    callers/tests that catch `IllegalArgumentException` are unaffected.
  - The remote-eval check in `RemoteEvalCoordinator` is retained as defense-in-depth;
    the single-context `GrowthBook` path keeps using `RemoteEvalOptionsValidator.validate(GBContext)`.
  - No new runtime dependencies. No public API removed.

### DIFF
--- a/lib/src/main/java/growthbook/sdk/java/exception/InvalidOptionsException.java
+++ b/lib/src/main/java/growthbook/sdk/java/exception/InvalidOptionsException.java
@@ -1,0 +1,39 @@
+package growthbook.sdk.java.exception;
+
+import lombok.Getter;
+
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.List;
+
+/**
+ * Thrown when GrowthBook options fail start-up validation.
+ *
+ * <p>Extends {@link IllegalArgumentException} so existing callers that catch invalid-argument errors
+ * keep working, while exposing the full list of problems via {@link #getViolations()} for callers
+ * that want structured access (diagnostics, health checks, structured logging).
+ */
+@Getter
+public class InvalidOptionsException extends IllegalArgumentException {
+
+    /**
+     * The individual, human-readable validation problems that were detected.
+     */
+    private final List<String> violations;
+
+    public InvalidOptionsException(String message, List<String> violations) {
+        super(message);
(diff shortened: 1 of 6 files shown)


madhuchavva (approved):
Approved. Startup option validation improves misconfiguration visibility and is consistent with JS/Python remote-eval guardrails. Focused OptionsValidator, RemoteEvalOptionsValidator, and GrowthBookClient tests passed locally.

PR #224, opened 2026-07-03
feat: Typed feature access

vazarkevych:
# Typed feature access (`FeatureKey` / `TypedKey`)

## Summary

Adds a type-safe way to reference feature flags so raw string keys stop leaking into call sites. Instead of:

```java
growthBook.isOn("new-home");
```

applications declare their flags once and reference them through a typed key:

```java
public final class Features {
    public static final TypedKey<Boolean> NEW_HOME  = TypedKey.ofBoolean("new-home");
    public static final TypedKey<String>  THEME     = TypedKey.ofString("theme");
    public static final TypedKey<Integer> MAX_ITEMS = TypedKey.ofInteger("max-items");
    private Features() {}
}

growthBook.getFeature(Features.NEW_HOME);        // FeatureResult<Boolean>
growthBook.getBooleanFeature(Features.NEW_HOME);
growthBook.getStringFeature(Features.THEME, "light");
growthBook.getIntegerFeature(Features.MAX_ITEMS, 10);
```

A mistyped key becomes a **compile-time error** rather than a silent runtime miss, and the value type is checked by the compiler (you can't call `getBooleanFeature` on a `String` key). The change is purely additive — every existing `String`-based method is untouched.

## Design

- **`FeatureKey<T>`** — an interface carrying both the feature identifier (`getKey()`) and its value type (`getValueType()`). Applications can implement it directly, or use the provided implementation.
- **`TypedKey<T>`** — an immutable implementation with factories: `of(key, Class)` for arbitrary Gson-deserializable types, plus `ofBoolean` / `ofString` / `ofInteger` / `ofDouble` / `ofFloat` shortcuts. Null-checked at construction; `equals`/`hashCode`/`toString` based on key + value type.
- **New overloads on both entry points** — every typed method mirrors an existing `String`-based one and simply delegates through `featureKey.getKey()` / `featureKey.getValueType()`, so all existing evaluation and type-conversion logic is reused:
  - `getFeature(FeatureKey<T>)` → `FeatureResult<T>`
  - `isOn(FeatureKey<?>)`, `isOff(FeatureKey<?>)`
  - `getFeatureValue(FeatureKey<T>, T defaultValue)` — infers the deserialization class from the key, so no separate `Class<T>` argument is needed
  - `getBooleanFeature` (with and without a default; default `false`), `getStringFeature`, `getIntegerFeature`, `getDoubleFeature`, `getFloatFeature`
- **`GrowthBook`** implements these via the shared package-private `IGrowthBook` interface (methods added there too). **`GrowthBookClient`** exposes the same set with a trailing `UserContext` parameter, each delegating through the typed `getFeatureValue` so the key's value type is the single source of truth.

## Behavior note

`getFeature(key)` mirrors `evalFeature`: `FeatureResult.getValue()` returns the **raw** evaluated value (a boxed primitive, or a `Map`/`List` for object and array features) — it is *not* deserialized into the key's value type. To get a deserialized instance of a complex type, use `getFeatureValue(FeatureKey<T>, T)`, which performs the Gson round-trip. This is documented on the methods.

## Testing

- **`TypedFeatureAccessTest`** (13 tests, single-context `GrowthBook`): `TypedKey` key/type exposure, null rejection, `equals`/`toString`; `getFeature` for present and unknown keys (`UNKNOWN_FEATURE` source, `isOff`); `isOn`/`isOff`; `getBooleanFeature` with/without default; all typed getters (string/int/double/float) reading values and falling back on unknown keys; complex-type deserialization via `getFeatureValue(TypedKey.of(key, MyConfig.class), …)`; a falsy feature reporting off.
- **`GrowthBookClientTest`** (multi-user path, real HTTP server): typed `getFeature`/`isOn`/`isOff`/`getBooleanFeature`/`getStringFeature`/`getIntegerFeature`/`getFeatureValue` plus numeric coercion for `getDoubleFeature`/`getFloatFeature` (including a whole-number JSON value `2` → `2.0f`) and unknown-key fallbacks.
- Full `./gradlew :lib:test` is green.

### DIFF
--- a/lib/src/main/java/growthbook/sdk/java/GrowthBook.java
+++ b/lib/src/main/java/growthbook/sdk/java/GrowthBook.java
@@ -16,6 +16,7 @@ import growthbook.sdk.java.exception.FeatureFetchException;
 import growthbook.sdk.java.model.AssignedExperiment;
 import growthbook.sdk.java.model.Experiment;
 import growthbook.sdk.java.model.ExperimentResult;
+import growthbook.sdk.java.model.FeatureKey;
 import growthbook.sdk.java.model.FeatureResult;
 import growthbook.sdk.java.model.GBContext;
 import growthbook.sdk.java.model.RequestBodyForRemoteEval;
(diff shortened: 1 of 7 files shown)


madhuchavva (approved):
Approved. Additive typed feature access helps avoid passing arbitrary strings at call sites. tests are good!

PR #225, opened 2026-07-06
ci: replace macOS runner with Linux in build matrix

vazarkevych:
The cross-platform CI workflow now runs on ubuntu-latest and windows-latest instead of macos-latest and windows-latest. The workflow file is renamed from ci_macos_windows.yml to ci_linux_windows.yml to match.


madhuchavva (approved):
Approved. This is a scoped CI runner change from macOS+Windows to Linux+Windows; no SDK runtime behavior changes.

Issue #226, opened 2026-07-15
Multi-user mode GrowthBookClient never enables SSE reconnection (retryOnFailure hardwired to false)


yang198876:
### Summary

The SSE reconnection support added for #52 (`GBFeaturesRepository.initialize(Boolean retryOnFailure)`) is never reachable from the multi-user mode `GrowthBookClient`: it always calls the no-arg `repository.initialize()`, which delegates to `initialize(false)`, and `Options` exposes no way to opt in. As a result, any app using `GrowthBookClient` with `FeatureRefreshStrategy.SERVER_SENT_EVENTS` permanently stops receiving feature updates after the first abnormal SSE stream failure — silently.

### Affected versions

Verified against the published sources of **0.10.6** and **0.10.10** (latest release), and against current `main`.

### Details (0.10.6 / 0.10.10)

The chain:

1. `GrowthBookClient.initialize()` → `repository.initialize()` (no-arg)
2. `GBFeaturesRepository.initialize()` → `initialize(false)` — `retryOnFailure` hardwired to `false`
3. In the SSE event source listener created by `createEventSourceListenerAndStartListening(retryOnFailure)`:

```java
@Override
public void onFailure(@NotNull EventSource eventSource, @Nullable Throwable t, @Nullable Response response) {
    super.onFailure(eventSource, t, response);
    if (retryOnFailure) {
        createEventSourceListenerAndStartListening(true);
        ...
    }
}
```

With `retryOnFailure == false`, an abnormal stream failure (connection reset, proxy/pod restart, mid-stream `IOException`) does **nothing**:

- no reconnect attempt;
- no polling fallback — `schedulePolling()` early-returns for `SERVER_SENT_EVENTS`;
- no observability — this path never invokes `onRefreshFailed(...)`, so registered `FeatureRefreshCallback.onError(...)` callbacks never fire.

The client then serves the last in-memory features forever (with `isCacheDisabled(true)` there is not even a cache to refresh from), and the application has no signal that anything is wrong.

Note that a *graceful* server close does reconnect on these versions (`onClosed` → `handler.onClose` → unconditional `createEventSourceListenerAndStartListening`), which makes the failure mode intermittent and easy to miss in testing: clean proxy restarts recover, abrupt failures don't.

### Still present on `main` (after #212)

[...]

PR #228, opened 2026-07-23
fix: enable SSE reconnect in multi-user client

vazarkevych:
## Problem

  In multi-user mode, `GrowthBookClient` always called the no-arg
  `repository.initialize()` (in `initializeFeaturesRepository()`), which delegates to
  `initialize(false)`. That hardwires `retryOnFailure = false`, so
  `scheduleSseReconnect(...)` early-returns and the SSE reconnect machinery is unreachable.

  As a result, a `GrowthBookClient` using `FeatureRefreshStrategy.SERVER_SENT_EVENTS`
  permanently stops receiving feature updates after the first abnormal stream failure
  (connection reset, proxy/pod restart, mid-stream IOException) — silently:

  - no reconnect attempt;
  - no polling fallback (`schedulePolling()` early-returns for SSE);
  - no observability — `FeatureRefreshCallback.onError(...)` never fires.

  Both `onFailure` and `onClose` route through `scheduleSseReconnect`, so multi-user + SSE
  also stops reconnecting on a *graceful* close. `Options.retryPolicy` (exponential
  backoff) was configurable but never actually engaged.

  ## Fix
  
  Expose the flag on `Options` and thread it through the single choke point:

  - `Options.sseReconnectOnFailure` — new option, **defaults to `true`**, with
    `isSseReconnectOnFailure()` returning `true` unless explicitly disabled. Added as the
    last constructor parameter to preserve the existing positional constructor signature.
  - `GrowthBookClient.initializeFeaturesRepository()` now calls
    `repository.initialize(options.isSseReconnectOnFailure())` instead of the no-arg form.

  Reconnect is on by default (bounded by `retryPolicy`), and callers can opt out with
  `Options.builder().sseReconnectOnFailure(false)`. The single-user
  `GBFeaturesRepository.initialize()` default (`false`) is unchanged to avoid a broader
  public-API behavior change.

  ## Tests
  backoff) was configurable but never actually engaged.

  ## Fix

  Expose the flag on `Options` and thread it through the single choke point:

  - `Options.sseReconnectOnFailure` — new option, **defaults to `true`**, with
    `isSseReconnectOnFailure()` returning `true` unless explicitly disabled. Added as the
    last constructor parameter to preserve the existing positional constructor signature.
  - `GrowthBookClient.initializeFeaturesRepository()` now calls
    `repository.initialize(options.isSseReconnectOnFailure())` instead of the no-arg form.

  Reconnect is on by default (bounded by `retryPolicy`), and callers can opt out with
  `Options.builder().sseReconnectOnFailure(false)`. The single-user
  `GBFeaturesRepository.initialize()` default (`false`) is unchanged to avoid a broader
  public-API behavior change.

### DIFF
--- a/lib/src/main/java/growthbook/sdk/java/multiusermode/GrowthBookClient.java
+++ b/lib/src/main/java/growthbook/sdk/java/multiusermode/GrowthBookClient.java
@@ -223,8 +223,6 @@ public class GrowthBookClient {
 
     private void initializeFeaturesRepository(GBFeaturesRepository repositorySnapshot) {
         try {
-            // Honor the configured SSE reconnect behavior; the no-arg initialize() hardwires
-            // retryOnFailure=false, which would disable SSE reconnection.
             repositorySnapshot.initialize(this.options.isSseReconnectOnFailure());
         } catch (FeatureFetchException e) {
             throw new GrowthBookClientInitializationException(
--- a/lib/src/main/java/growthbook/sdk/java/multiusermode/GrowthBookClient.java
+++ b/lib/src/main/java/growthbook/sdk/java/multiusermode/GrowthBookClient.java
@@ -223,7 +223,9 @@ public class GrowthBookClient {
 
     private void initializeFeaturesRepository(GBFeaturesRepository repositorySnapshot) {
         try {
-            repositorySnapshot.initialize();
+            // Honor the configured SSE reconnect behavior; the no-arg initialize() hardwires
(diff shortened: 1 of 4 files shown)


yang198876:
Hi @vazarkevych, thank you again for preparing this fix for #226.

When you have a chance, could this PR please be reviewed and, if everything looks good, included in the next release? We would really appreciate it. Thank you!

